PR #5, opened 2023-02-23
PyTest runner in github action

jwestw:
# Summary

I have created a github action yaml file in the `.github/workflows` folder. This should cause pytest to be run whenever there is a PR - which I am testing now by creating a PR. :)

What this github action should do:
 - using miniconda and python 3.6
 - run pytest to run all tests in the tests folder (there is only one at the moment)
 - gets a coverage report which displays on the console (in the github action)

I recognise that the coverage report being hidden in the console of the github action runner is not that useful, but all requirements of 63 and 64 are met here. 

I will create another ticket to have the coverage report put out into a github comment or onto the readme

# This closes the following:

RDRP-63
RDRP-64

# Checklists

This pull/merge request meets the following requirements:

- [x] code runs
- [x] [developments are ethical][data-ethics-framework] and secure
- [x] you have made proportionate checks that the code works correctly
- [x] test suite passes
- [x] developments adhere to AQA plan (see `docs/aqa/aqa_plan.md`)
- [x] data log updated (see `docs/aqa/data_log.md`), if necessary **N/A**


tests/test_main.py
```diff
@@ -0,0 +1,18 @@
+"""Create a test suite for the main module."""
+
+import pytest
+
+def test_add():
+    """Test the add function."""
```
ali-i3, line 6: Use the standard AAA (Arrange-Act--Assert) pattern for unit tests

tests/test_main.py
```diff
@@ -0,0 +1,18 @@
...
+    from src.utils.testfunctions import add
+    # use pytest to assert the result
+    assert add(1, 2) == 3
+    assert add(10, 222) == 232
```
ali-i3, line 10: Is this second test necessary

ali-i3 (changes requested):
It looks great! 
Pytest runs on my local machine
GitHub actions also runs
Export the coverage report

tests/test_main.py
jwestw: Yes I think so

ali-i3 (approved):
This all looks great!

PR #7, opened 2023-03-07
52 toml parser

RoddyMacrae:
# Summary

This uses the toml python package (https://pypi.org/project/toml/)  so in order for it to work you need to install toml (using pip currently instead of poetry):

pip install toml

Once this done you can then run the main.py from the src directory as normal.

Deliverables: 

1. Config file should be in the `config` folder in the in the root, so that users are aware of where to easily access it
2. Should have (working) documentation for users on how to use config file - documentation to be updated each time a new parameter is added

Both completed.
Below bullet points outline other aspects done while completing the ticket.

- Updated the toml file to include more relevant data
- Created a toml parser function that returns a dictionary
- Uncommented the 'run_pipeline()' function
- Corrected the add import location from 'src.utils.testfunctions' to 'utils.testfunctions' which made it run

# Checklists

This pull/merge request meets the following requirements:

- [x] code runs
- [ ] [developments are ethical][data-ethics-framework] and secure
- [ ] you have made proportionate checks that the code works correctly
- [ ] test suite passes
- [ ] developments adhere to AQA plan (see `docs/aqa/aqa_plan.md`)
- [ ] data log updated (see `docs/aqa/data_log.md`), if necessary
- [ ] assumptions, and caveats log updated (see `docs/aqa/assumptions_caveats.md`), if
  necessary
- [ ] [minimum usable documentation][agilemodeling] written in the `docs` folder

Comments have been added below around the incomplete checks.

[agilemodeling]: http://agilemodeling.com/essays/documentLate.htm
[data-ethics-framework]: https://www.gov.uk/government/publications/data-ethics-framework


RoddyMacrae:
The new commits fix the issue where toml_parser() was returning the a print function and the docstrings not being correctly formatted/descriptive. It also fixes the issue where the toml file path was not correctly pointing to the file.

toml_parser now returns a dictionary of the parsed userconfig.toml file. The docstring is now in line with section 3.8.3 of the style guide found here: https://google.github.io/styleguide/pyguide.html

src/main.py
```diff
@@ -1,7 +1,25 @@
...
+        If only one variable under the section title in the TOML file is given
+        then it is passed directly as a dictionary value. If more than one
+        variable is defined then they are parsed as a dictionary themselves.
+        An example of what is returned is given below:
+        
+        {'title': 'TOML Example config', 'period': {'start_period': 
+        datetime.date(1990, 10, 10), 'end_period': datetime.date(2000, 10, 5)}, 
+        'source_file': {'location': 'D:/Data', 'fileName': 'file.txt'}, 'output_location': 
+        {'hive_db': 'hive.db', 'tableName': 'name.table'}, 'outlier_correction': 
+        {'location': 'D:/', 'fileName': 'outliers.txt', 'bool': True}}
```
jwestw, line 20: This is good, but maybe the example doesn't need to be so long.

src/main.py
```diff
@@ -1,7 +1,25 @@
...
 
-from src.utils.testfunctions import add
+import toml
+from utils.testfunctions import add
 
+def toml_parser():
+    """Function to parse the userconfig.toml file.
+
+    Returns:
+        A dictionary where the keys are section titles within the TOML file.
```
jwestw, line 10: A nested dictionary, I think

src/main.py
```diff
@@ -1,7 +1,25 @@
...
+        variable is defined then they are parsed as a dictionary themselves.
+        An example of what is returned is given below:
+        
+        {'title': 'TOML Example config', 'period': {'start_period': 
+        datetime.date(1990, 10, 10), 'end_period': datetime.date(2000, 10, 5)}, 
+        'source_file': {'location': 'D:/Data', 'fileName': 'file.txt'}, 'output_location': 
+        {'hive_db': 'hive.db', 'tableName': 'name.table'}, 'outlier_correction': 
+        {'location': 'D:/', 'fileName': 'outliers.txt', 'bool': True}}
+    """    
+    return toml.load("/home/cdsw/research-and-development/config/userconfig.toml")
```
jwestw, line 22: The path `"config/userconfig.toml"` also works. No need for absolute paths.

src/main.py
```diff
@@ -1,7 +1,25 @@
 """The main pipeline"""
 
-from src.utils.testfunctions import add
+import toml
+from utils.testfunctions import add
 
+def toml_parser():
```
jwestw, line 6: Thinking about this function name, is it descriptive enough? Will this parse all toml files, or just one in particular? Could the name be more exact/descriptive and give clues about what it's actually doing?

src/main.py
jwestw, line 22: I personally prefer more explicit code (but I understand there are no hard rules on this). Since we are trying to make our code as readable as possible for the rest of the team let's make it more explicit

e.g.

```python
some_parameter = "spam_eggs"
object_to_return = some_process(some_parameter)

return object_to_return
```

jwestw, line 22: Let's define the path outside of the function, then feed the path to the function as a parameter. Usually I would like the path defined in a config file but this is a necessary exception.

jwestw (changes requested):
A few more changes and I think we are done with this.

src/main.py
jwestw: `user_config_path = "config/userconfig.toml"` should do it.

jwestw (changes requested):
Please make the changes according to the comments I've given. 

Once this is moved to `helpers.py`, I think this ticket will be done then.

src/main.py
RoddyMacrae: Good point. Have resolved this in latest commit.

src/main.py
RoddyMacrae: I agree

src/main.py
jwestw: Maybe it will be a general purpose toml parser, if so, great, let's keep it like that. In that case the docstring would need to reflect its more general function

src/main.py
RoddyMacrae: Good point. I have fixed this in latest commit.

RoddyMacrae: Done

src/main.py
RoddyMacrae: Would something like user_config_reader be suitable?

RoddyMacrae:
Addressed conflict in main.py and helpers.py.

jwestw (approved):
Brilliant. Thanks for resolving all my queries.

jwestw:
I also just resolved a few more conflicts.

PR #10, opened 2023-03-14
76 Updating the pull request template

jwestw:
Please both review the new pull request template. We need to all agree on what is included there so we are all working to the same standards.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,34 +1,72 @@
-# Summary
+## Pull Request submission 
 
-Add your summary here - keep it brief, to the point, and in plain English. [For further
-information about pull requests, check out the GDS
-Way](https://gds-way.cloudapps.digital/standards/pull-requests.html).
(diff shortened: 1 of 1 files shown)


ali-i3 (changes requested):
This is looking fantastic!

I have a few ideas i got from reading through this [guide](https://best-practice-and-impact.github.io/qa-of-code-guidance/peer_review.html)

For the person submitting you could add:
- A check on if any new high level parameters are put in dedicated configuration files
- Check if the latest version of the develop branch is integrated into the working branch (is a rebase necessary before merge?)

For the reviewer you could add:
- A checkbox to see if the pull request warrants a discussion on upgrading the version.
- The reviewer generates the same outputs from running the code (to avoid possibility of successful run with different results)

RoddyMacrae (commented):
This looks really great. 

I can't think of any further suggestions other than what Ilyas has suggested currently.

ali-i3 (approved):
Looks great!

PR #19, opened 2023-04-25
119 file present

RoddyMacrae:
## Pull Request submission 

- Created a function to check if a file exists and is non-empty in src/data_validation/validation.py
- Function returns a boolean. True if file exists and non-empty, otherwise False.
- Relative path to file is passed to the function, current default is set to the validation.py file itself so default returns True.
- Created a unit test in tests/test_validation.py.
- Unit test covers True result, False result and False result where file exists but is empty.
- Imported src.data_validation.validation into main.py and tested the code ran, it did succesfully, however I had to comment out logger code as there were some import errors regarding Config_settings and csv_creator. Branch was created from default, unsure if there are known issues or not regarding modules not existing.

*Insert any instructions to help the reviewer, e.g. "install new requirements from `environment.yml`"*
- Comment out logger code in src/main.py to test import of src/data_validation/validation.py works correctly. I think issue stems from different merges not quite being finished on default. 
 
- check_file_exists() defaults to validation.py so will return true. You can pass any file path to it as an argument. Unit test creates its own file and should delete it after it is used for testing.

#### Closes or fixes

Closes #

#### Code

- [ x] **Code runs** The code runs on my machine and/or CDSW
- [ x] **Conflicts resolved** There are no conflicts (I have performed a rebase if necessary)
- [ x] **Requirements** My/our code functions according to the requirements of the ticket
- [ x] **Clean Code**
    - [ x] Code is as [PEP 8]([url](https://peps.python.org/pep-0008/)) compliant as I can humanly make it
    - [ x] Code passess flake8 linting check
    - [ x] Code adheres to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)
- [ x] **Type hints** All new functions have [type hints ](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html)

#### Documentation

- [ x] **Function Documentation** Docstrings within the function(s')/methods have been created
    - [ x] Includes `Args` and `returns` for all major functions 
    - [ x] The docstring details data types

#### Data
- [ x] Secrets checker pre-commit passes

#### Testing
- [ x] **Unit tests** Unit tests have been created and are passing _or a new ticket to create tests has been created_

---

# Peer Review Section

---

### Review comments

  works correctly?)


jwestw (changes requested):
Hi Roddy, thanks for the thorough pull request submission. 

This ticket was to check that the **data** file is present and non-empty and is part of our work on the Data Validation and Correction module. Your code does not check that a data file is present, it checks that a python file is present. It does not make sense to do this as we would never want to check the presence of the Python file (as this would always be True). 

Could you default the path towards a data folder? Perhaps the `.gitkeep` file will do as a default if you feel you need to provide one, but I see no need to have a default file. 

Making  part of the path default probably makes sense though - so `data` or `data/raw` for example. 

Please also make a note to bring up the issues you've been having with the logger in a team scrum. This may be a bug that needs resolving.

src/data_validation/validation.py
```diff
@@ -14,10 +14,11 @@ def check_file_exists(filePath="./src/data_validation/validation.py") -> bool:
     """
     output = False
 
-    fileExists = os.path.exists(filePath)
+    fileExists = os.path.exists(filepath + filename)
```
jwestw, line 17: This looks really good. Have you tried it with a little data file?

src/data_validation/validation.py
```diff
@@ -0,0 +1,24 @@
...
+
+    Returns:
+        A bool: boolean value is True if file exists and is non-empty,
+        False otherwise.
+    """
+    output = False
+
+    fileExists = os.path.exists(filepath + filename)
+    if fileExists:
+        fileSize = os.path.getsize(filepath + filename)
```
jwestw: `filepath + filename` may work but I believe it's better practice to use `os.path.join(filepath, filename)` because it works across different os's. However if it works in Dev/Test then no need to change it.

src/data_validation/validation.py
jwestw: I think it belongs in validation.

src/data_validation/validation.py
jwestw: @RoddyMacrae could you please make the suggested change or respond?

tests/test_validation.py
```diff
@@ -0,0 +1,32 @@
...
+    # developer_config.yaml should exist and be non-empty
+    result_1 = check_file_exists("developer_config.yaml", "./src/")
+    result_2 = check_file_exists("Non_existant_file.txt")
+    result_3 = check_file_exists(empty_file.name)
+
+    # Delete emptyfile.py after testing
+    os.remove(empty_file.name)
+
+    # Assert
+    assert isinstance(result_1, bool)
```
jwestw: This just checks if it is a boolean, not if it's either `True` or `False`. Is this what you want?

src/data_validation/validation.py
jwestw: @RoddyMacrae I don't think you've responded to this

tests/test_validation.py
jwestw: @RoddyMacrae I don't think you've responded to this. Please take action!

RoddyMacrae: Yes that is what I intended with this. Perhaps there is a better practice RE assertion testing. I can always add an assert True/False in place of this?

src/data_validation/validation.py
RoddyMacrae: This is a good point. I have amended the changes.

src/data_validation/validation.py
```diff
@@ -48,8 +43,6 @@ def get_masterlist(masterlist_path) -> pd.Series:
     return masterlist
 
 
-@time_logger_wrap
-@exception_wrap
```
jwestw: These wrappers should not be getting removed.

src/data_validation/validation.py
```diff
@@ -104,8 +97,6 @@ def validate_post_col(df: pd.DataFrame, masterlist_path: str) -> bool:
             f"Invalid postcodes found: {combined_invalid_postcodes.to_list()}"
         )
 
-    validationlogger.info("All postcodes validated....")
```
jwestw, line 107: Not sure why this is getting removed.

jwestw (approved):

PR #57, opened 2023-08-15
instance fix

jwestw:
Just fixing the problem with the sub-responder toml missing "instance"

### DIFF
--- a/config/subresponder_schema.toml
+++ b/config/subresponder_schema.toml
@@ -1,15 +1,3 @@
-[instance]
-Description = "nan"
-Deduced_Data_Type = "Int64"
-Nullable = "nan"
-Current_Data_Type = "object"
-Length = "nan"
-Min_values = "nan"
-Max_values = "nan"
-Possible_Categorical_Values = ["nan"]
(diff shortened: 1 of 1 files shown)


config/subresponder_schema.toml
```diff
@@ -1,3 +1,15 @@
...
+Deduced_Data_Type = "Int64"
+Nullable = "nan"
+Current_Data_Type = "object"
+Length = "nan"
+Min_values = "nan"
+Max_values = "nan"
+Possible_Categorical_Values = ["nan"]
+q_code = ""
+old_ref = "nan"
+
```
AnneONS: There are already two versions of ['instance'] in the subresponder_schema, so one needs to be removed, not added.

AnneONS (dismissed):
That's fine, great.

AnneONS (commented):
Just one thing, I've noticed that this [instance] contains 
q_code = ""
old_ref = "nan"
which isn't in the other items in this schema.

ali-i3 (approved):
Looks good!

PR #74, opened 2023-08-25
389 validation bug fix

dpurches:
## Pull Request submission

Addressed two issues:

1. `nan` string values were appearing in the output `.csv` files. This was due to a really weird quirk of of the `.astype` function in pandas. If you apply `df[col].astype('str')` to a column with strings and np.nan in then this will convert the np.nan values to the string 'nan' and the column dtype will remain as 'object'. However if you do the very similar command of `df[col].astype('string')` on this column then the np.nan values remain as actual nan values (so won't be written as a string in the output csv) but also the column dtype changes to 'string' (which I think is nullable). This is quite odd (and maybe a bit confusing) behaviour and I'm not quite sure why it happens. I have addressed this by updating the `full_responses` validation to apply the second function (above) when 'str' is specified in the config.
2. There was also a bug in the pipeline where the merge on the manual outliers file was failing. This was due to the type of the `instance` column in the manual outliers df being `float` when it was read in as a `.csv`. I have applied the validation function after it has been read in using a new toml config file for this input.

#### Closes or fixes

* RDRP-389

#### Code

- [x] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/staging/validation.py
```diff
@@ -211,7 +211,7 @@ def validate_data_with_schema(survey_df: pd.DataFrame, schema_path: str):
     }
 
     # Cast each column individually and catch any errors
-    for column in survey_df.columns:
+    for column in dtypes_dict.keys():
```
dpurches, line 215: This was changed as the `toml` file for the manual outliers does not contain details for all the columns - only the ones we actually use. Does this logic make sense?

src/staging/spp_parser.py
```diff
@@ -25,8 +25,8 @@ def parse_snap_data(snapdata: dict) -> Tuple[pd.DataFrame, pd.DataFrame]:
     responsesdict = snapdata["responses"]
 
     # Make dataframes
-    contributors_df = pd.DataFrame(contributordict)
-    responses_df = pd.DataFrame(responsesdict)
+    contributors_df = pd.DataFrame(contributordict).replace("nan", None)
+    responses_df = pd.DataFrame(responsesdict).replace("nan", None)
```
AnneONS: Nice solution! Was this all that's needed to get rid of those "nans"?

src/staging/validation.py
```diff
@@ -275,12 +275,14 @@ def combine_schemas_validate_full_df(
             # Convert non-integer string to NaN
             survey_df[column] = survey_df[column].apply(pd.to_numeric, errors="coerce")
             # Cast columns to Int64
-            survey_df[column] = survey_df[column].astype(pd.Int64Dtype())
+            survey_df[column] = survey_df[column].astype("Int64")
```
AnneONS, line 279: Seems a good change to me, it works!

config/manual_outliers_schema.toml
```diff
@@ -0,0 +1,29 @@
...
+Nullable = false
+Current_Data_Type = "str"
+Length = "nan"
+Min_values = "nan"
+Max_values = "nan"
+Possible_Categorical_Values = ["nan"]
+
+[manual_outlier]
+Description = "nan"
+Deduced_Data_Type = "bool"
```
ali-i3: The "bool" data-type converts all strings (including empty cells) as True. This gives us 15000 manual outliers which is what alerted us. I got this working using "boolean" instead because it's nullable. 

This is a similar problem to "str" vs "string" datatypes.

src/developer_config.yaml
```diff
@@ -37,6 +38,7 @@ network_paths:
   outliers_path: "R:/BERD Results System Development 2023/DAP_emulation/outliers"
   manual_outliers_path: "R:/BERD Results System Development 2023/DAP_emulation/outliers/manual_outliers/manual_outlier_2023-08-22_v45.csv"
```
ali-i3: Can you change this file to "manual_outlier_2023-08-29_v67.csv" which is a much more recent one with a more accurate amount of outliers?

ali-i3 (changes requested):
This looks great! There's only 2 small changes I could think of and I put comments on the lines. These would fix the problems me and Anne were having.

src/staging/validation.py
AnneONS: Actually, this isn't needed, as we're only reading in the columns from the toml file, and so by the time we validate we only have the 3 columns. Keeping in the survey_df.columns lets us check actual cols we read in are correct, and will give us better info in the error logging as we'll get the name of the column in the actual dataframe (rather than in the toml)

AnneONS (commented):
With the changes Ilyas suggested (and changing back to survey_df.columns in validation.py) it seems to be working well!

config/manual_outliers_schema.toml
dpurches: Good spot, have updated this now.

src/staging/spp_parser.py
dpurches: Tbh I'm not sure we still need this with what I updated using the config file - might check and get rid of if not.

src/staging/validation.py
dpurches: I've just had a look and I think the columns we specify to be read in don't come from the toml file (they do match but are just hard coded). We could either make sure these are read from the toml file (just so we don't accidentally change one but not the other) or keep the code above (this isn't as refined). What do you think?

src/staging/validation.py
dpurches: Just fyi, I don't think this change actually has any impact - `astype(pd.Int64Dtype()) <=> astype('Int64')`

I'll just keep it as a string as it's a bit more readable.

ali-i3 (approved):
Looks good!

PR #76, opened 2023-08-30
344 try

zorge69:
## Pull Request submission

*Changed the way the path to config file is used*
The path is hardcoded in main.py
In pipeline.py/run_pipeline, an object Config_settings is called with config_path as an argument
Platform-specific libraries are imported now inside run_pipeline function
Config is passed to all modules and functions as a dictionary


main.py
```diff
@@ -5,6 +5,7 @@
 
 # reload the pipeline module to implement any changes
 reload(src)
+config_path = r"D:\repos\research-and-development\src\developer_config.yaml"
```
ali-i3: The root of this path ("D:\repos\") is not the same for all of us. The pipeline fails unless everyone who runs it renames their root folder to the same thing. This might cause further issues with git and conda setup so I need to test this further.

tests/test_staging/test_validation.py
```diff
@@ -11,11 +12,15 @@
     validate_data_with_schema,
     combine_schemas_validate_full_df,
 )
-from src.utils.helpers import Config_settings
+
 
 # Get the config
-conf_obj = Config_settings()
-config = conf_obj.config_dict
+def generate_config(val):
```
ali-i3, line 17: I've made some minor changes to the tests that used the previous config. It's all fixed now.

ali-i3 (commented):
This looks good! There is a pipeline breaking issue regarding the path to the config that will affect all of us and we all need to discuss this before we move forward. Apart from this, everything else works as expected.

AnneONS (commented):
This all seems to work fine, and I think it will be ready to merge when we have decided what to do about us all having different setups on our D drive (meaning the config file path is different for all of us)

ali-i3 (approved):
Looks good! All conflicts are resolved now and pipeline runs as expected.

PR #78, opened 2023-09-04
399 creating short form output and schema

ali-i3:
## Pull Request submission

Creates a short form output dataframe with the columns specified by the team.

All short form questions that were in the estimation module have been replaced in the output by the estimated values.

The SIC numbers are used in the staging module to create a new column 'product_group' since this is a many-to-one mapping and there are no nulls in the SIC column.

A schema is present with the current names of the columns and the required names for future cross reference.

To see the output, take a look at the file: `R:\BERD Results System Development 2023\DAP_emulation\outputs\output_short_form\output_short_form2023-09-04_v80.csv`

#### Closes or fixes

Closes [RDRP-398](https://jira.ons.gov.uk/browse/RDRP-398) & [RDRP-399](https://jira.ons.gov.uk/browse/RDRP-399)

#### Code

- [x] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


jwestw:
I have just run this code, it ran end-to-end no problem. Just checking the code now.

src/developer_config.yaml
```diff
@@ -2,15 +2,14 @@ global:
   log_to_file: True # Write logs to .log file
   logging_level: "DEBUG"
   table_config: "SingleLine"
-  validate_postcodes: True
```
jwestw, line 5: Why has this been removed? Is it a mistake?

src/outputs/map_output_cols.py
```diff
@@ -0,0 +1,125 @@
...
+        df.loc[conditions[i], "sizeband"] = decisions[i]
+
+    # Convert datatype to int
+    df["sizeband"] = df["sizeband"].astype("Int64")
+
+    return df
+
+
+def create_cora_status_col(df, mapper_df, main_col="statusencoded"):
+    """_summary_
```
jwestw: This docstring needs updating/creating.

src/outputs/outputs_main.py
```diff
@@ -37,17 +36,29 @@ def run_output(
...
+    estimated_df = map.map_sizebands(estimated_df)
+
+    # Map the itl regions using the postcodes
+    estimated_df = map.join_itl_regions(estimated_df, postcode_itl_mapper)
+
+    # Prepare the shortform output dataframe
+    short_form_df = short.run_shortform_prep(estimated_df, round_val=4)
+
+    # Create short form output dataframe with required columns from schema
+    schema_path = "src/outputs/output_schemas/frozen_shortform_schema.toml"
```
jwestw: Should this path be in the config file?

src/outputs/short_form_out.py
```diff
@@ -96,12 +92,51 @@ def run_shortform_prep(
...
 
     # create columns for headcounts for civil and defense
     df = create_headcount_cols(df, round_val)
 
     return df
+
+
+def create_shortform_df(df: pd.DataFrame, schema: str) -> pd.DataFrame:
+    """Creates the shortform dataframe for outputs with
+    the required columns. The naming
```
jwestw: "The naming"... this needs finishing. :)

src/outputs/short_form_out.py
```diff
@@ -96,12 +92,51 @@ def run_shortform_prep(
...
+        output_schema[column_nm]["old_name"]: column_nm
+        for column_nm in output_schema.keys()
+    }
+
+    # Create empty dataframe for outputs
+    output_df = pd.DataFrame()
+
+    # Add required columns from schema to output
+    for key in colname_dict.keys():
+        output_df[colname_dict[key]] = df[key]
```
jwestw: I think I can see a couple more ways to do this, one more concisely. Firstly, don't forget the Pandas `DataFrame` constructor has a `columns` parameter. What could you provide at this stage?

Alternatively, if you still want to do it in two stages, consider `.rename` for the second stage which also takes a `columns` argument. You could avoid looping and making iterative changes to a dataframe which is always good! :)

src/staging/pg_conversion.py
```diff
@@ -111,10 +130,51 @@ def sic_to_pg_mapper(
         )
     # Map to the target column using the dictionary taking into account the null values.
     # Then convert to categorigal datatype
-    df[sic_column] = pd.to_numeric(df[sic_column], errors="coerce")
-    df[target_col] = df[sic_column].map(map_dict)
-    df[target_col] = df[target_col].astype("category")
+    filtered_df[sic_column] = pd.to_numeric(filtered_df[sic_column], errors="coerce")
+    filtered_df[target_col] = filtered_df[sic_column].map(map_dict)
+    filtered_df[target_col] = filtered_df[target_col].astype("category")
```
jwestw, line 135: Good to see the datatype change here.

jwestw (dismissed):
A few minor changes (mostly doctstrings) need actioning. Otherwise it's all looking good.

src/developer_config.yaml
ali-i3: This config would skip the entire postcodes module. We need this to run to get the harmonised postcodes for use later in the pipeline. I think we only had this setting since it used to take really long to run.

I think we should maybe also remove the "postcode_csv_check" setting since it returns an empty list when false. What do you recommend?

src/outputs/outputs_main.py
ali-i3: Good point. We don't have any of our toml schemas in our developer config yet, so i've added now a new section called schema_paths.

src/outputs/short_form_out.py
ali-i3: I assumed the order of the output columns mattered so i assigned columns according to the schema order. If they don't matter then I can use an intersection and rename it like this in 2 steps:

    # Create subset dataframe with only the required outputs
    output_df = df[df.columns.intersection(colname_dict.keys())]
    
    # Rename columns to match the output specification
    output_df.rename(columns={key :colname_dict[key] for key in colname_dict},inplace = True)

If the order matters, then I can add this third step:
    
    # Rearrange to match user defined output order
    output_df = output_df[colname_dict.values()]

smithv1 (approved):
Code looks good, runs fine and tests pass. Moved D:\RandD\research-and-development\tests\test_staging\test_pg_conversion.py from test_imputation to be in line with repo folder structure

smithv1:
Code runs fine with new changes and unit tests pass

AnneONS (approved):
Checked all with Ilyas

PR #81, opened 2023-09-07
Postcode fix 2

ali-i3:
## Pull Request submission

After a conversation with Ana today, we realised we are checking the wrong column for postcodes. This is a quick fix that changes the column from "referencepostcode" to "601". There a fewer errors in the postcodes now. (21 in total). I also fixed the tests and ran the pre-commits.

Changed postcode check column to 8 digit 'pcd2' column

Added a postcodes_harmonised column for future itl mappers.

#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


smithv1 (approved):
Code changes look good, code runs on main and tests pass

dpurches:
@dpurches to review once new column added.

src/staging/validation.py
```diff
@@ -79,8 +85,26 @@ def validate_post_col(
             f"Invalid pattern postcodes found: {invalid_pattern_postcodes.to_list()}"
         )
```
dpurches: Aware it's not part of the code you've added, but think it would be good to log here when we don't find any invalid postcodes as well - arguably a bit more important as this is more likely to be a mistake.

Rather than an `if/else` maybe we just log the invalid postcodes we find (regardless of whether or not we find any)

src/staging/validation.py
```diff
@@ -58,10 +62,12 @@ def validate_post_col(
     if not isinstance(df, pd.DataFrame):
         raise TypeError(f"The dataframe you are attempting to validate is {type(df)}")
 
-    # Check if postcodes match pattern
-    # Create list of postcodes with incorrect patterns
-    invalid_pattern_postcodes = df.loc[
-        ~df["referencepostcode"].apply(validate_postcode_pattern), "referencepostcode"
+    # Can only validate not null postcodes
+    val_df = df.loc[~df["601"].isnull()]
```
dpurches, line 63: Running through this with the debugger, the minority of values in this column are non-null (533/37,535). Is this to be expected or just happens because we are using test data?

src/staging/validation.py
```diff
@@ -58,10 +62,12 @@ def validate_post_col(
...
+
+    # Apply the pattern validation function
+    invalid_pattern_postcodes = val_df.loc[
+        ~val_df["601"].apply(validate_postcode_pattern), "601"
```
dpurches, line 67: I think there may be a more efficient way of doing this, but given we are working with the invalid postcodes (presumably a small dataset) probably not worth trying to work out how.

src/staging/validation.py
```diff
@@ -79,8 +85,26 @@ def validate_post_col(
             f"Invalid pattern postcodes found: {invalid_pattern_postcodes.to_list()}"
         )
 
+    # Remove the invalid pattern postcodes from the masterlist test
+    val_df = df.loc[~df.index.isin(invalid_pattern_postcodes.index.to_list())]
```
dpurches, line 88: Just an thought for if the data was bigger, this might be more efficient with a join/merge. As with my comment above, probs not worth changing here but maybe one for the future!

src/staging/validation.py
```diff
@@ -79,8 +85,26 @@ def validate_post_col(
...
+
+    # Clean and harmonise the "601" postcodes to match the masterlist
+    check_real_df = clean_postcodes(val_df, "601")
+
+    # Only validate not null postcodes
+    check_real_df = check_real_df.loc[~check_real_df["601"].isnull()]
```
dpurches, line 94: I think it would be good to add a comment in the code to distinguish between the two sections - first we validate the pattern and then second we validate that the postcodes are real (I think!)

src/staging/validation.py
```diff
@@ -79,8 +85,26 @@ def validate_post_col(
...
+    unreal_postcodes = check_pcs_real(df, check_real_df, postcode_masterlist, config)
+
+    # Create empty column for harmonised postcodes
+    df["postcodes_harmonised"] = np.nan
+
+    # Return harmonised postcodes for all rows of postcodes
+    harmonised_df = clean_postcodes(val_df, "postcodes_harmonised")
+    df.loc[harmonised_df.index, "postcodes_harmonised"] = harmonised_df[
+        "postcodes_harmonised"
+    ]
```
dpurches:
```suggestion
    df.loc[harmonised_df.index, "postcodes_harmonised"] = clean_postcodes(val_df, "postcodes_harmonised")["postcodes_harmonised"]
```
Not sure if this suggestion is valid - could also change the output of `clean_postcodes` to be a column rather than a df.

src/staging/validation.py
```diff
@@ -133,7 +158,64 @@ def get_masterlist(postcode_masterlist) -> pd.Series:
     return masterlist
 
 
-def check_pcs_real(df: pd.DataFrame, postcode_masterlist: pd.DataFrame, config: dict):
+def clean_postcodes(df, column):
+    """Clean the postcodes to fit the masterlist format"""
```
dpurches: Could do with a bigger docstring to explain in more detail the steps/results. eg:
```
 - remove whitespace of 8+ characters
 - add whitespace for 8 digit postcodes
 etc etc
```

src/staging/validation.py
```diff
@@ -79,8 +85,26 @@ def validate_post_col(
...
+
+    # Only validate not null postcodes
+    check_real_df = check_real_df.loc[~check_real_df["601"].isnull()]
+
     # Create a list of postcodes not found in masterlist
-    unreal_postcodes = check_pcs_real(df, postcode_masterlist, config)
+    unreal_postcodes = check_pcs_real(df, check_real_df, postcode_masterlist, config)
+
+    # Create empty column for harmonised postcodes
+    df["postcodes_harmonised"] = np.nan
```
dpurches, line 100: I think this sets the column type to `float64` which is then converted to `object` when the postcodes are added. Will that be fine later on or do we need to make sure this is a string column?

dpurches (approved):
Looks good! Runs fine and all the tests pass as well.

Most of the comments are pretty minor - feel free to challenge me on any of them!

smithv1 (approved):
Code runs fine and test pass, can probably leave doc strings as is for now

src/staging/validation.py
ali-i3: Yep, there's only 533 postcodes in "601" after transposition. Only for some long forms with instance > 1. This looks right

src/staging/validation.py
ali-i3: Done. It now logs the invalid postcodes even when empty. I've also added a line logging the number of instances of invalid and not real postcodes.

src/staging/validation.py
ali-i3: I run in to all sorts of indexing issues with the suggested method since the 2 don't match in size/shape. The current method does it in 2 steps but works fine

src/staging/validation.py
ali-i3: This works fine since I can account for none types during the mapping later in outputs. The remainder of the column are strings so this is fine

src/staging/validation.py
dpurches: Yeh that makes sense - keep it as is then, probably slightly more readable as well at present.

dpurches:
Looks good - happy for you to merge 😀

PR #83, opened 2023-09-11
413 keep lastupdateddate col (staging fix)

AnneONS:
## Pull Request submission

Feedback from the testers (see Jira ticket 413) identified the following:

1) The business area wants column `lastupdateddate` to be retained, this is now not dropped from the contributors dataframe. 

This takes place in `staging/spp_snapshot_processing.py` and the corresponding test has been updated. 

NOTE: by default, the datatype for lastupdateddate, datetime64, writes to Excel in a strange, unreadable format. For this reason, I've added a `date_format` parameter to `to_csv` in the `write_local_csv` and `write_hdfs_csv` functions, please have a look.

2) The columns `frozensicoutdated`, `rusicoutdated`, `frozensic`, `rusic` were missing leading zeros in Excel. 
We considered ensuring these were strings in the pipeline, but this could cause conflicts with mapping to SIC codes where these are read in from csv (in which case they might not contain the leading zero). So we decided to keep these values as integers.
So no change here.

#### Closes or fixes

Closes #413

#### Code

- [x] **Clean Code**

#### Documentation

#### Data

#### Testing
- [x] **Unit tests** Unit tests have been created and are passing _or a new ticket to create tests has been created_

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/utils/local_file_mods.py
```diff
@@ -44,7 +44,7 @@ def write_local_csv(filepath: str, data: pd.DataFrame):
     # Open the file in write mode
     with open(filepath, "w", newline="\n", encoding="utf-8") as file:
         # Write dataframe to the file
-        data.to_csv(file, index=False)
+        data.to_csv(file, date_format="%Y-%m-%d %H:%M:%S.%f+00", index=False)
```
smithv1, line 47: Tested by add;
         data["timestamp"] = pd.to_datetime(data["timestamp"], format="%d/%m/%Y-%H:%M:%S")
         data["timestamp"] = data["timestamp"].dt.strftime("%m/%d/%Y-%H:%M:%S")

And got an empty data error - not sure if there's a better way to test this

smithv1 (commented):
Code looks good, test pass and runs fine. Would like to check if there's another way to test the code: "date_format="%Y-%m-%d %H:%M:%S.%f+00", index=False)"

src/utils/local_file_mods.py
AnneONS: Many thanks for looking at this. I chose this date_format by looking at the `Description = "Datetime format = %Y-%m-%d %H:%M:%S.%f+00"` in  the `contributors_schema.toml`. I tested the output looked good in the full_responses output, and noted that it was the same format as in the original (old) version of the contributors csv given by the business area.

smithv1: Ran with and without and got expected output so good to merge this one now

smithv1 (approved):

PR #87, opened 2023-09-13
353 154 manifest file

jwestw:
## Pull Request submission

- Added a lot of paths for hdfs as these were missing
- Fixed the config path in `main/py` so it works on hdfs as well as windows
- Added a print out of the time for the whole pipeline (this can be removed if not needed but it's good for developers to know)
- Added a `dev_test` switch in the config
- Fixed a few paths in the `network_paths` section of the config
- created a (basic) output schema for the short_form output file
- integrated the manifest code from the cis_households project into our project
- adapted the hdfs related functions such as `hdfs_write_string_to_file` and `hdfs_isfile` etc
- created the equivilent functions for local file system
- made sure the manifest class is working
- fixed many bugs

#### Closes or fixes

closes RDRP-353
closes RDRP-154

#### Code

- [ ] **Code runs** The code runs on my machine and/or CDSW
- [ ] **Conflicts resolved** There are no conflicts (I have performed a rebase if necessary)
- [ ] **Requirements** My/our code functions according to the requirements of the ticket
- [ ] **Dependencies** I have updated the environment yaml so it includes any new libraries I have used
- [ ] **Configuration file updated** any high level parameters that the user may interact with have been put into the config file (and imported to the script)
- [ ] **Clean Code**
    - [ ] Code is as [PEP 8]([url](https://peps.python.org/pep-0008/)) compliant as I can humanly make it
    - [ ] Code passess flake8 linting check
    - [ ] Code adheres to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)
- [ ] **Type hints** All new functions have [type hints ](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html)

#### Documentation

Any new code includes all the following forms of documentation:

- [ ] **Function Documentation** Docstrings within the function(s')/methods have been created
    - [ ] Includes `Args` and `returns` for all major functions
    - [ ] The docstring details data types
- [ ] **Updated Documentation**: User and/or developer working doc has been updated

#### Data
- [ ] All data needed to run this script is available in Dev/Test
- [ ] All data is excluded from this pull request
- [ ] Secrets checker pre-commit passes

#### Testing
- [ ] **Unit tests** Unit tests have been created and are passing _or a new ticket to create tests has been created_

---

# Peer Review Section

- [ ] All requirements install from (updated) `environment.yaml`
- [ ] Documentation has been created and is clear - **check the working document**
[...]

PR #91, opened 2023-09-21
355 outgoing files

jwestw:
## Pull Request submission

This PR should contain all code changes for ticket RDRP-355 and also those for RDRP-153 and RDRP-353 (the latter two were contained in PR #87, but this branch was made from that one, so should contain all changes in PR #87)

**Main change**: The script moves the file(s) the user has selected and creates the manifest file (single) containing the metadata for the selected (potentially multiple) csv files. 

Changes in this PR:
- Created a stand-alone pipeline for exporting data which is initiated from export_main.py which is in the root directory
- Most of the script to export files is in `export_files.py` and there is a `run_export` function which ties everything together to export the files
- Created a `get_file_choice` function which relies on user input to gather which files they want to export 
- There were some modifications to the manifest writing class, such as checking that the file is not more than 2.5GB
- Uses Ilyas's new short form schema to validate the outgoing file columns
- renamed the folder from "outgoing" to "outgoing_export"

Changes from PR #87 

- Added a lot of paths for hdfs as these were missing
- Fixed the config path in `main/py` so it works on hdfs as well as windows
- Added a print out of the time for the whole pipeline (this can be removed if not needed but it's good for developers to know)
- Added a `dev_test` switch in the config
- Fixed a few paths in the `network_paths` section of the config
- created a (basic) output schema for the short_form output file
- integrated the manifest code from the cis_households project into our project
- adapted the hdfs related functions such as `hdfs_write_string_to_file` and `hdfs_isfile` etc
- created the equivilent functions for local file system
- made sure the manifest class is working
- fixed many bugs

#### Closes or fixes

closes RDRP-353
closes RDRP-154

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/outputs/manifest_output.py
```diff
@@ -0,0 +1,202 @@
...
+            )
+        if not isdir_func(export_directory):
+            raise ManifestError(
+                f"Outgoing directory does not exist: {self.outgoing_directory}"
+            )
+
+        if not isinstance(pipeline_run_datetime, datetime):
+            raise ManifestError("Pipeline run datetime must be a datetime object.")
+
+        self.manifest_datetime = pipeline_run_datetime.strftime("%Y%m%d_%H%M%S")
```
ali-i3: The output file name is hard to read (20230928_171230_manifest.json). Can you change the format of the datetime slightly to make it more readable especially for users?

src/outputs/export_files.py
```diff
@@ -0,0 +1,177 @@
...
+else:
+    OutgoingLogger.error("The network_or_hdfs configuration is wrong")
+    raise ImportError
+
+OutgoingLogger.info(f"Using the {network_or_hdfs} file system as data source.")
+
+
+# Define paths
+NETWORK_OR_HDFS = config["global"]["network_or_hdfs"]
+paths = config[f"{NETWORK_OR_HDFS}_paths"]
```
ali-i3: You've converted this to lower case in other places. Maybe the same applies here

src/outputs/outputs_main.py
```diff
@@ -3,12 +3,20 @@
...
 import src.outputs.short_form_out as short
 import src.outputs.map_output_cols as map_o
 
 OutputMainLogger = logging.getLogger(__name__)
 
+# Get the shortform schema
+short_form_schema = toml.load("src/outputs/output_schemas/frozen_shortform_schema.toml")
+
+# Columns are needed for filtering
+short_form_cols = short_form_schema.keys()
```
ali-i3: This code doesn't get called anywhere. Are these short form cols meant to be here?

src/outputs/outputs_main.py
```diff
@@ -27,14 +35,17 @@ def run_output(
...
         cora_mapper (pd.DataFrame): used for adding cora "form_status" column
 
 
     """
+
     OutputMainLogger.info("Starting short form output...")
 
     NETWORK_OR_HDFS = config["global"]["network_or_hdfs"]
-    output_path = config[f"{NETWORK_OR_HDFS}_paths"]["output_path"]
+    paths = config[f"{NETWORK_OR_HDFS}_paths"]
```
ali-i3, line 43: Lower case?

ali-i3 (commented):
This code looks incredible! I've put in a partial review since I haven't tested this in dev/test. I've ran export_main.py a few times and outputted a few files to the outgoing folder. I've put in some minor comments but please test this further.

src/estimation/calculate_weights.py
```diff
@@ -98,6 +98,8 @@ def calculate_weighting_factor(
 
     for cell_number, cell_group in groupd_by_cell:
 
+        # cell_number += 1 TODO: investigate if this is needed in HDFS
+
```
AnneONS, line 99: Yes, there's a key error as it's looking for key 0, don't know why.

Does this cell_number += 1 need to be conditional on dev_test == True, or will it work for both?

AnneONS: An alternative would be to follow what we do in TMI (though this hasn't been tested in devtest)

`    grp = filtered_df.groupby("imp_class")`
`    class_keys = list(grp.groups.keys())`

`    for var in target_variable_list:`
`        for k in class_keys:`
`            # Get subgroup dataframe`
`            subgrp = grp.get_group(k)`

AnneONS (commented):
Impressive code James! 
Just done a light review at the moment, and commented on `calculate_weights.py`.

The pipeline fails on updating the runlog, couldn't initially see how to correct this.

Have you recently pulled in develop?

src/developer_config.yaml
```diff
@@ -22,22 +22,26 @@ runlog_writer:
     display: False # Display the runlog in the terminal
     log_path: "data/log"
 hdfs_paths:
-  root: None
-  logs_foldername: "/ons/rdbe_dev/logs/run_logs"
-  snapshot_path: "/ons/rdbe_dev/snapshot-202012-002-fba5c4ba-fb8c-4a62-87bb-66c725eea5fd.json"
+  root: "/ons/rdbe_dev/"
```
AnneONS, line 26: this needs to be updated to:
`  logs_foldername: "/ons/rdbe_dev/logs/run_logs"`

AnneONS (commented):
I've added a correction in the comments for the developer_config, the pipeline now runs in DevTest.

By the way, do we need the export_files functionality on the Network, or was this for testing?

src/outputs/manifest_output.py
jwestw: This is a difficult one. I agree that's not particularly user friendly but I need it to be something that is unique. As the manifest can contain metadata for many csv files, I can't use a single file name as the basis for the manifest file name.

src/outputs/export_files.py
jwestw: Fixed

src/outputs/outputs_main.py
jwestw: Fixed. It was orphan code

src/estimation/calculate_weights.py
jwestw: Thanks for the fix @AnneONS

src/developer_config.yaml
jwestw: Thanks for the fix

src/outputs/export_files.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Returns:
+        selction_list (list): A list of the files to transfer."""
+    file_list = list_files(output_path)
+    choices = []
+    print("Select a file to transfer:")
+    for i, file in enumerate(file_list):
+        print(f"{i+1}. {file}")
+    file_num = "X"  # initialize file_num to a non-integer value
+    while True and file_num != "Q":
```
AnneONS: I tried entering 'q', and realised my mistake shall we make it "Q" or "q" so it's easier on the user?

src/outputs/manifest_output.py
AnneONS: how about YYYY-MM-DD_etc ?

AnneONS (commented):
a couple more comments!

src/outputs/manifest_output.py
jwestw: I have updated the name. I kept in there the time because it makes it more unique. Having the time in there with a ":" in was causing the bug in hdfs though! I think it's sorted now. 

Also, I don't think the manifest file needs to be very user-readable, as it's for nifi to pick up.

jwestw:
I think everything is resolved now.

AnneONS (approved):
Tested this thoroughly in DevTest and happy it works as expected

PR #97, opened 2023-10-06
311 tmi step 2

AnneONS:
## Pull Request submission

TMI Step 2
New script `src.imputation.impute_civ_def.py` created to impute R&D type (civil or defence in  q200)

"C" for civil or "D" for defence is imputed to q200 from randomly based on their proportions in imputation classes.

Unit tests were created for the calculation of proportions for the imputation classes.

#### Closes or fixes

Closes #311

#### Code

- [ ] **Clean Code**

#### Documentation

- [x] **Function Documentation** Docstrings within the function(s')/methods have been created

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/imputation/impute_civ_def.py
```diff
@@ -0,0 +1,258 @@
...
+
+    Returns:
+        Dict[str, Tuple(float, float)]
+    """
+    # create dictionary to hold civil or defence ratios for each class
+    pgsic_dict = {}
+
+    # Filter out imputation classes that are missing either "201" or "rusic"
+    # and exclude empty pg_sic classes
+    cond1 = ~(df["pg_sic_class"].str.contains("nan")) & (df["empty_pgsic_group"] == False)
```
dpurches: Can we not use `pd.isNull(df['pg_sic_class'])` for this condition - would this be safer?

src/imputation/tmi_imputation.py
```diff
@@ -290,7 +325,7 @@ def create_mean_dict(df, target_variable_list):
 
     # Filter out imputation classes that are missing either "200" or "201"
     filtered_df = filtered_df[~(filtered_df["imp_class"].str.contains("nan"))]
-    filtered_df = filtered_df[filtered_df["formtype"] == formtype_long]
+    # filtered_df = filtered_df[filtered_df["formtype"] == formtype_long]
```
dpurches: Can this be removed?

src/imputation/impute_civ_def.py
```diff
@@ -0,0 +1,259 @@
...
+            to use in imputation based on product group and SIC.
+        pg_dict (Dict[str, Tuple(float, float)]): Dictionary with proportions
+            to use in imputation based on product group only.
+
+    Returns:
+        pd.DataFrame: The same dataframe with R&D type imputed.
+    '"""
+    df["200_original"] = df["200"]
+    df["200_imputed"] = df["200"]
+    df["200_imp_marker"] = "N/A"
```
dpurches: Should this be 'nan' or `np.nan` for consistency maybe?

AnneONS: good question! Actually, following what we have in TMI step 3, I actually want the string "N/A" as in "not applicable". It comes out as the string "N/A" in excel, but if it's ambiguous because it could be interpreted as a null in some contexts, we could possibly change it to "no_imputation"

src/imputation/impute_civ_def.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # Create logic conditions for filtering
+    clear_status_cond = df["statusencoded"].isin(["210", "211"])
+    to_impute_cond = (
+        ((df["status"] == "Form sent out") | (df["604"] == "No"))
+        & df["instance"] != 0
```
dpurches: Could we just filter to remove `instance == 0` at a higher level? I.e. in the `impute_civil_defence` function.

src/imputation/impute_civ_def.py
dpurches: Yeh I think 'no_imputation' may be a bit clearer - I don't know if pandas might interpret these as `nan` in some scenarios.

AnneONS: added this, also in tmi_imputation

src/imputation/impute_civ_def.py
AnneONS: Good shout, @dpurches But to join again, do you think it's more efficient to use .update() or to use .concat followed by sort_by("reference", "instance") ?

dpurches: My hunch would be `.update` as it's just one method, but not entirely sure. Tbh your code is pretty efficient atm so not too worried either way 😃

dpurches (approved):
Happy to approve! All unit tests pass, code looks good and happy with the output.

Couple of observations (not necessarily issues) from the output:

 - For the vast majority where we impute a value of 'D', we are using fall back imputation. In fact there is only one class where we use the `pgsic` group to impute.
 - In this one group, there is only one real instance and this has a value of `D` - presumably meaning the proportions will be 100% D. Do we have any threshold for the minimum number of true values in a class for us to deem it worth using the proportions from there?

Both not a blocker for merging, just wanted them to be raised.

src/imputation/impute_civ_def.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def calc_cd_proportions(df: pd.DataFrame):  # -> Tuple[float, float]:
+    """Calc the proportion of civil and defence entries
+
+    The proportions are calculated for a dataframe representing a
+    single imputation class, which contains at least one non-null
+    entry for column 200, so division by zero will be avoided.
+    """
+    num_civ = len(df.loc[df["200"] == "C", "200"])
```
smithv1, line 20: Would these ever be zero leading to denominator zero?

src/imputation/impute_civ_def.py
```diff
@@ -0,0 +1,248 @@
...
+    # create imputation classes based on product group and rusic
+    df = tmi.create_imp_class_col(df, "201", "rusic", "pg_sic_class")
+
+    # flag empty pg_sic_classes in new bool col "empty_pgsic_class"
+    df = calc_empty_group(df, "pg_sic_class", "empty_pgsic_group")
+
+    # create a new imputation class called "pg_class" based on product group
+    # (col 201) only, to be used when imputation cannot be performed using
+    # both prodcut group and SIC.
+    df = tmi.create_imp_class_col(df, "201", None, "pg_class")
```
smithv1, line 118: Wondering about using 201 or df["pg_class"] = df["201]

tests/test_imputation/test_impute_civil_defence.py
```diff
@@ -0,0 +1,211 @@
...
+        assert out_d1 == 0.375
+
+        out_c2, out_d2 = calc_cd_proportions(input_df2)
+
+        assert out_c2 == 1.0
+        assert out_d2 == 0.0
+        
+
+class TestCreateCivdefDict:
+    """Unit tests for calc_cd_proportions function."""
```
smithv1: exchange func name for create_civdef_dict

tests/test_imputation/test_impute_civil_defence.py
```diff
@@ -0,0 +1,211 @@
...
+
+        pgsic_dict, pg_dict = create_civdef_dict(input_df)
+
+        exp_pg_dict = {"AC":(0.6, 0.4)}
+
+        assert pg_dict == exp_pg_dict
+
+
+class TestPrepCDImpClasses:
+    """Unit tests for calc_cd_proportions function."""
```
smithv1: exchange for func name prep_cd_imp_classes

smithv1 (approved):

tests/test_imputation/test_impute_civil_defence.py
AnneONS: Good point

src/imputation/impute_civ_def.py
AnneONS: No, it won't! See comment on lines 17 and 18

src/imputation/impute_civ_def.py
AnneONS: That nearly works but not quite, as it won't cover the cellno 817 special case.

tests/test_imputation/test_impute_civil_defence.py
AnneONS: will do.

PR #111, opened 2023-10-27
RDRP-483 - Carry forward

dpurches:
## Pull Request submission

 - Implements the first part of the Mean of Ratios method - carrying forward clean data from the previous years data into `form sent out` or `clear - check needed` statuses in the current data.

- please check the output of impution_qa against the develop branch, for an up-to-date output of the develop branch, see imputation_qa/full_responses_imputed_2023-11-01_v42.csv

## TODO: 
Still need to map numeric PG to alpha-numeric.
 
#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


config/backdata_schema.toml
```diff
@@ -0,0 +1,70 @@
...
+
+[status]
+Description = "nan"
+Deduced_Data_Type = "category"
+Nullable = false
+Current_Data_Type = "str"
+Length = ">=1"
+Min_values = "nan"
+Max_values = "nan"
+Possible_Categorical_Values = ["Clear", "Clear - overridden", "Form saved", "Clear - overridden SE", "Form sent out", "Check needed", "Combined child (NIL2)", "Out of scope (NIL3)", "Ceased trading (NIL4)", "Dormant (NIL5)", "Part year return (NIL8)", "No UK activity (NIL9)"]
```
AnneONS: Clear - overridden SE - this one is a red herring, can be removed if we're doing other changes.

src/imputation/MoR.py
```diff
@@ -0,0 +1,85 @@
+"""Functions for the Mean of Ratios (MoR) methods."""
+import pandas as pd
+import re
+
+from src.imputation.tmi_imputation import apply_to_original
+from src.imputation.apportionment import run_apportionment
+
+good_statuses = ["Clear", "Clear - overridden"]
+bad_statuses = ["Form sent out", "Check needed"]
```
AnneONS, line 8: I like the way you've used this for good and bad!

src/imputation/MoR.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Args:
+        df (pd.DataFrame): Processed full responses DataFrame
+        backdata (pd.DataFrame): One period of backdata.
+        target_vars ([string]): List of variables to impute.
+
+    Returns:
+        pd.DataFrame: df with MoR applied
+    """
+    to_impute_df, backdata = mor_preprocessing(df, backdata)
```
AnneONS: I like the way you've followed conventions in our other code with naming, that's helpful.

src/imputation/MoR.py
```diff
@@ -0,0 +1,85 @@
...
+    Args:
+        df (pd.DataFrame): full responses for the current year
+        backdata (pd.Dataframe): backdata file read in during staging.
+    """
+    # Select only values to be imputed and remove duplicate instances
+    to_impute_df = df.copy().loc[
+        (
+            (df["formtype"] == "0001")
+            & (df["status"].isin(bad_statuses))
+            & (df["instance"] == 0 | pd.isnull(df["instance"]))
```
AnneONS: The only way we'd get a null instance is for "form sent out" and in TMI we create these as instance 1 (long forms only)
I suggest we add in a TODO here to move this to imputation_main.

tests/test_imputation/test_impute_civil_defence.py
```diff
@@ -78,31 +73,213 @@ def create_input_df(self):
...
 
         data = [
-            [1001, 0, np.nan, np.nan, 0, "Clear", "800", "1234", "nan_1234", False, "nan", False],
-            [1001, 1, "C", "AC", 100, "Clear", "800", "1234", "AC_1234", False, "AC", False],
-            [1001, 2, "C", "AC", 200, "Clear", "800", "1234", "AC_1234", False, "AC", False],
-            [1001, 3, "D", "AC", 50, "Clear", "800", "1234", "AC_1234", False, "AC", False],
-            [1002, 0, np.nan, np.nan, 0, "Clear", "800", "1234", "nan_1234", False, "nan", False],
-            [1002, 1, "C", "AC", 100, "Clear", "800", "1234", "AC_1234", False, "AC", False],
-            [2002, 0, np.nan, np.nan, np.nan, "Clear", "800", "444", "nan_444", False, "nan", False],
-            [2002, 1, np.nan, "AC", 200, "Clear", "800", "444", "AC_444", True, "AC", False],
```
AnneONS: Please could we reverse this to how it was before? The pre-commit has totally messed this up so it's not possible to read! Can we remove the line length check for tests?

AnneONS (commented):
Still reviewing, looks good so far!

src/imputation/imputation_main.py
```diff
@@ -43,6 +45,8 @@ def run_imputation(
     # Apportion cols 4xx and 5xx to create FTE and headcount values
     df = run_apportionment(df)
 
+    df = run_mor(df, backdata, target_vars)
+
```
AnneONS: this needs to go after this code on line 53
df["imp_marker"] = "no_imputation"

Also, could it go after the code on lines 55-58, so we put the carry forward values in the _imputed cols?

**important**

We need to update the TMI code so it doesn't do TMI where CF or MoR have happened

src/imputation/MoR.py
```diff
@@ -0,0 +1,85 @@
...
+        df (pd.DataFrame): Processed full responses DataFrame.
+        backdata (_type_): One period of backdata.
+        target_vars (_type_): Variables to be imputed.
+
+    Returns:
+        pd.DataFrame: df with values carried forwards
+    """
+    df = pd.merge(df, backdata, how="left", on="reference", suffixes=("", "_prev"))
+    for var in target_vars:
+        df[var] = df.loc[:, f"{var}_prev"]
```
AnneONS: instead of this, could we have 
 df[f"{var}_imputed"] = df.loc[:, f"{var}_prev"]

src/imputation/MoR.py
```diff
@@ -0,0 +1,85 @@
...
+        target_vars ([string]): List of variables to impute.
+
+    Returns:
+        pd.DataFrame: df with MoR applied
+    """
+    to_impute_df, backdata = mor_preprocessing(df, backdata)
+
+    # Carry forwards method
+    carried_forwards_df = carry_forwards(to_impute_df, backdata, target_vars)
+    carried_forwards_df["imp_marker"] = "CF"
```
AnneONS: This is getting overridden in imputation_main.py

Also we need to update TMI so it doesn't do TMI in these cases, or we'll lose the changes.

AnneONS (commented):
we need to change the code in a few places

update tmi_imputation to not impute where MoR

config/backdata_schema.toml
dpurches: Sure, was just copied from the wide responders so may not make sense.

tests/test_imputation/test_impute_civil_defence.py
dpurches: Good shout - I'll revert the commit and make sure it doesn't affect the tests.

src/imputation/MoR.py
```diff
@@ -0,0 +1,98 @@
...
+    Args:
+        df (pd.DataFrame): full responses for the current year
+        backdata (pd.Dataframe): backdata file read in during staging.
+    """
+    # TODO move this to imputation main
+    # Select only values to be imputed and remove duplicate instances
+    imputation_condition = (
+        (df["formtype"] == "0001")
+        & (df["status"].isin(bad_statuses))
+        & (df["instance"] == 0 | pd.isnull(df["instance"]))
```
AnneONS: for instance 0 there are no values for the target variables, instance 0 holds metadata and in long forms the target variables appear on the other instances. Also, instance 0 doesn't have "C" and "D" values so the don't belong to an imputation class. Could you work with instance 1 instead?

dpurches: That might be a good idea - we're using this to pull out the bad statuses in the data which we then go on to overwrite some of using carry forwards. However at the moment this does mean we'd lose stuff going into TMI.

AnneONS: I think that we want to use MoR as the first priority, so we want to do as much as possible with MoR instead of TMI (which isn't supposed to be as good). So we'd want to pull out all the baddies and do as many as possible with MoR Does that make sense with how you're thinking of it?

dpurches: Yeh I think so - the reason I filter by instance is that we only have unique references in the current data when we join to the backdata. That way the pattern of instance is carried over from the backdata. In terms of the metadata in instance one does that make sense?

AnneONS: Ah yes, so for carried forward you're not using imputation classes, so of course, just keeping instance 0 makes sense.

dpurches: Although when we come to add MoR we will need imputation classes so don't know if that means we should use instance 1?

AnneONS: We need to include ALL instances in the to_impute_df and then filter for instance 0 later. Otherwise we've got duplicates of instance 1 etc with CF in one of them and TMI in the other.

src/imputation/tmi_imputation.py
```diff
@@ -367,11 +367,12 @@ def apply_tmi(df, target_variables, mean_dict):
     filtered_df = filter_by_column_content(
         df, "status", ["Form sent out", "Check needed"]
     )
-
     # Filter out any cases where 200 or 201 are missing from the imputation class
     # This ensures that means are calculated using only valid imputation classes
     # Since imp_class is string type, any entry containing "nan" is excluded.
     filtered_df = filtered_df[~(filtered_df["imp_class"].str.contains("nan"))]
+    # Only apply TMI to values that are yet to be imputed
+    filtered_df = filtered_df[filtered_df["imp_marker"] == "no_imputation"]
```
AnneONS: This can be removed- I've split the df at the very start of TMI so that TMI steps 2 and 4 don't apply to the CF/ MoR rows

coatet:
Reviewed, code looks good and pipeline runs through on my end!

JenCheshire (approved):
Looks good!

src/staging/staging_main.py
```diff
@@ -253,6 +253,48 @@ def run_staging(
...
+        backdata = None
+        StagingMainLogger.info("Loading of Backdata File skipped")
+
+    # # Load the PG mapper
+    # pg_mapper = paths["pg_mapper_path"]
+    # check_file_exists(pg_mapper)
+    # pg_mapper = read_csv(pg_mapper)
+
+    # # Map PG from SIC/PG numbers to column '201'.
+    # full_responses = pg.run_pg_conversion(full_responses, pg_mapper, target_col="201")
```
AnneONS: @JenCheshire can you check this is correct to have this commented out?

AnneONS: OK, I've sorted this.

PR #115, opened 2023-11-01
RDRP-489, RDRP-490, RDRP-491: Construction module

coatet:
## Pull Request submission

- Implement `run_construction()` function to launch the construction module
- Implement `get_amendments()` and `get_additions()` to get changes from updated snapshot
- Implement `output_construction_files()` to save two construction files for users to approve of constructed records
- Implement `apply_construction()`, `apply_amendments()` and `apply_additions()` to read user-approved construction files and apply them to the main snapshot data
- Modify `pipeline.py` to run the construction module and pass the resulting dataframe to the imputation module, including logic if construction is turned off

This code runs locally using the network drive but because there are no common keys between the 2020 snapshot and the 2022 snapshot, it doesn't produce an amendments file. As a result, the amendments code hasn't been tested, and can't be without using a valid updated snapshot, which is only in prod. 

#### Closes or fixes

* Closes RDRP-489, RDRP-490 and RDRP-491

#### Code

- [x] **Clean Code**

#### Documentation

#### Data
- [x] All data is excluded from this pull request

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/staging/staging_main.py
```diff
@@ -216,7 +216,10 @@ def run_staging(
             os.path.join(feather_path, f"{snapshot_name}_responses.feather")
         )
         full_responses = snapdata
-        secondary_full_responses = secondary_snapdata
+        if load_updated_snapshot:
+            secondary_full_responses = secondary_snapdata
+        else:
+            secondary_full_responses = None
```
AnneONS: Thanks for fixing this one! :-)

src/construction/construction.py
```diff
@@ -1,16 +1,243 @@
...
+            and added from the construction files. amended
+    """
+    # Skip this module if the secondary snapshot isn't loaded
+    load_updated_snapshot = config["global"]["load_updated_snapshot"]
+    if load_updated_snapshot is False:
+        construction_logger.info("Skipping Construction...")
+        return main_snapshot
+
     # ! For now, we add the year column since neither file has it
     main_snapshot["year"] = 2022
```
AnneONS, line 47: there's some code that does this in outputs/short_form.py - you could call the func or use this:
    df["period_year"] = df["period"].astype("str").str[:4]

(period_year is the col name for the short form outputs, would it work to use this in your code too, and we could just add it in during staging?

AnneONS (commented):
This isn't a full review, I just noticed some things trying to run my branch.

src/construction/construction.py
```diff
@@ -1,16 +1,243 @@
...
+        try:
+            additions_df = read_csv(additions_filepath)
+            constructed_df = apply_additions(main_df, additions_df)
+        except pd.errors.EmptyDataError:
+            construction_logger.warning(f"Additions file {additions_filepath} is empty, skipping...")
+
+    # Save the constructed dataframe as a CSV
+    tdate = datetime.now().strftime("%Y-%m-%d")
+    # ? Should this go in a subfolder of the outputs folder?
+    construction_output_filepath = os.path.join(paths["output_path"], f"construction_amendments_{tdate}_v{run_id}.csv")
```
JenCheshire: The filename for the amendments construction file (line 151) is the same as the output filename for the constructed df: I think this one needs changing

src/construction/construction.py
```diff
@@ -1,16 +1,243 @@
...
+
+    construction_logger.info(f"{accepted_amendments_df.shape[0]} records amended during construction")
+
+    return amended_df
+
+
+def apply_additions(main_df, additions_df):
+    """Apply additions to the main snapshot."""
+    # Drop records where accept_changes is False and if any remain, add them to main df
+    accepted_additions_df = additions_df.drop(additions_df[~additions_df.accept_changes])
```
JenCheshire: Getting a KeyError here: think it needs accepted_additions_df = additions_df.drop(additions_df[~additions_df.accept_changes]**, axis=1**)

src/construction/construction.py
```diff
@@ -1,16 +1,243 @@
...
+        except pd.errors.EmptyDataError:
+            construction_logger.warning(f"Amendments file {amendments_filepath} is empty, skipping...")
+
+    if additions_exist:
+        try:
+            additions_df = read_csv(additions_filepath)
+            constructed_df = apply_additions(main_df, additions_df)
+        except pd.errors.EmptyDataError:
+            construction_logger.warning(f"Additions file {additions_filepath} is empty, skipping...")
+
```
JenCheshire, line 193: Once the additional records are added, the indexing is causing issues in imputation - this sorted it: constructed_df = constructed_df.sort_values(["reference", "instance"], ascending=[True, True]).reset_index(drop=True)

JenCheshire (commented):
Nearly finished reviewing - will do the rest first thing tomorrow.  Looking good so far, just a couple of comments

src/construction/construction.py
JenCheshire: Ignore the above!  Just relooked at this, and that wasn't the problem - it was dropping all the records, even if accept_changes was set to TRUE.  Tweaked it to: accepted_additions_df = additions_df.drop(additions_df[~additions_df["accept_changes"]].index) and it has run and the additional records are showing in the final constructed df.  
However, I'm now getting a 'ZeroDivisionError: float division by zero' in imputation

src/construction/construction.py
JenCheshire: Having relooked at the accepted_changes comment above, this might not be issue

src/construction/construction.py
coatet: Did you mark one of the first few rows with `accept_changes` true? The first few records don't have anything in the 202 column and it looks like the zero division is happening with an operation in 202 as well. I'm doing a run with some rows that have something in 202 marked to accept changes to see if that's it.

JenCheshire: Ah, yes - I just marked the first 13 as true

JenCheshire: I had everything pointed at R:\BERD Results System Development 2023\DAP_emulation\construction, so you can see my outputs (construction_final_2023-11-03_v74 is the constructed df at the end of construction with additional records - I used construction_additions_2023-11-02_v50  as my additions file, with top 13 records marked as true)

coatet: Just ran it and it gives the same error even when they both have 202s. But the `calc_fte_column()` function is using a list of new and old columns for its loop - I know there's some differences between the columns in the 2022 data and the 2020 snapshot I'm using as the fake updated snapshot, so it might be that? It's hard to diagnose since I don't know what that function is for!

src/construction/construction.py
coatet: Good catch, renamed this.

src/construction/construction.py
coatet: Resolving as this seems to be caused by the zero division bug above

src/construction/construction.py
coatet: After a bunch of testing I can't find any reason for this. The function divides by `tot_202_CD` but this column is nonzero when the error occurs. At the moment I'm fairly confident this is an issue with the fake updated snapshot I have to use outside of DAP.

dpurches (approved):

PR #121, opened 2023-11-08
486 more imputation fixes

AnneONS:
## Pull Request submission

This branch implements several fixes for the imputation module:

* Expansion imputation (step 4 TMI long forms): where there are zeros in the target variables, the breakdown cols were nulls. This is fixed by removing division by zero and entering zeros for the breakdown cols instead.
* MoR carry forwards: removed "postcode only" rows from the backdata before merging, also resetting the index at the end of the process.
* Census imputation for short forms: methodology advised that "P" selection entries should not after all be included in the calculation of values to impute.
* Short form expansion imputation: removed division by zero error to ensure breakdown columns are not null. 
* Short form expansion imputation: include trimmed short form rows in the calculations and in the evaluations

#### Closes or fixes

Closes #486

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/imputation/MoR.py
```diff
@@ -29,7 +29,7 @@ def run_mor(df, backdata, impute_vars):
     carried_forwards_df = carry_forwards(to_impute_df, backdata, impute_vars)
 
     # TODO Remove the `XXX_prev` columns (left in for QA)
-    return pd.concat([remainder_df, carried_forwards_df])
+    return pd.concat([remainder_df, carried_forwards_df]).reset_index()
```
dpurches: @AnneONS just a heads up this is the bugfix I added. I think it was caused by us not doing anything with the index when we performed the merge.

AnneONS: Great, thanks.

src/imputation/MoR.py
```diff
@@ -56,8 +53,12 @@ def mor_preprocessing(df, backdata):
     backdata = backdata.rename(columns=to_rename)
 
     backdata = run_apportionment(backdata)
+
+    clear_status_cond = backdata["status"].isin(good_statuses)
+    # identify backdata rows that only consist of postcodes to eliminate this
+    postcode_only_cond = backdata["211"].isnull() & backdata["405"].isnull()
     # Only pick up useful backdata
-    backdata = backdata.loc[(backdata["status"].isin(good_statuses)), :]
+    backdata = backdata.loc[clear_status_cond & ~postcode_only_cond, :]
```
AnneONS, line 61: @dpurches this is the fix so we don't get a load of empty rows added with carry forwards.

dpurches: Makes sense to me!

src/imputation/MoR.py
dpurches: @coatet This should solve the index issue

AnneONS: @dpurches  is there a way of solving the problem without it creating an extra "index" column? This comes in the output csv even though we are writing with "index=False"

AnneONS: Don't worry, I googled it :-) it's reset_index(drop=True)

src/imputation/expansion_imputation.py
```diff
@@ -18,38 +18,46 @@ def evaluate_imputed_ixx(
...
+    # Return the groupby object unaltered if there are no TMI values
+    # in the imputation class
+    TMI_mask = group["imp_marker"] == "TMI"
+    imputed_subgroup = group.loc[TMI_mask]
+
+    if imputed_subgroup.empty:
+        ExpansionLogger.debug(
+            f"Imputation class {imp_class} has no TMI rows to process."
+        )
+        return group
```
dpurches, line 34: This will break the execution of this function - is that what we want to do?

src/imputation/imputation_main.py
```diff
@@ -35,9 +35,11 @@ def run_imputation(
     # Convert shortform responses to longform format
     df = run_short_to_long(df)
 
-    # TODO: imp_marker should be 'R' for clear responders
-    # Initialise imp_marker column, default value "no_imputation"
-    df["imp_marker"] = "no_imputation"
+    # Initialise imp_marker column with a value of 'R' for clear responders
+    # and a default value "no_imputation" for all other rows for now.
+    clear_responders_mask = df.status.isin(["Clear", "Clear - overridden"])
```
dpurches, line 40: Maybe not one for right now, but just generally I think it would be good to store the clear repsonders codes at a higher level and maybe in a config .

src/imputation/expansion_imputation.py
```diff
@@ -18,38 +18,46 @@ def evaluate_imputed_ixx(
     """Evaluate the imputed 2xx or 3xx as the sum of all 2xx or 3xx
     over the sum of all 211 or 305 values, multiplied by the imputed 211.
     """
+    imp_class = group["imp_class"].values[0]
+    ExpansionLogger.debug(f"Imputation class: {imp_class}.")
+    ExpansionLogger.debug(f"Master column: {master_col}.")
+
+    # Return the groupby object unaltered if there are no TMI values
+    # in the imputation class
+    TMI_mask = group["imp_marker"] == "TMI"
```
dpurches, line 27: I think I'm just missing this, but where do we set the values in the `imp_marker column`?

src/imputation/expansion_imputation.py
```diff
@@ -18,38 +18,46 @@ def evaluate_imputed_ixx(
...
 
-        # Update the imputation column for status encoded 100 and 201
-        # i.e. for non-responders
-        unclear_statuses = ["Form sent out", "Check needed"]
-        unclear_mask = group["status"].isin(unclear_statuses)
-        imputed_value = (sum_breakdown_q / sum_master_q) * master_imp_val
-        # Write imputed value to the non-responder records
-        group.loc[unclear_mask, f"{bd_col}_imputed"] = imputed_value
+        # Update the imputed break-down columns where TMI imputation has been applied
+        group.loc[TMI_mask, f"{bd_col}_imputed"] = imputed_value
```
dpurches, line 60: This seems a lot cleaner! Are the 0s we're assigning here wrapped into the question Ana had about do we impute with 0s?

dpurches (approved):
All looks good! Mainly just questions from me. I haven't run it, I can do, but also happy for you to just merge yourself if you're happy with the outputs

src/imputation/imputation_main.py
AnneONS: Maybe... but having them in the config would mean passing the config through lots of funcs. I was thinking of putting this line we use so much in a function in a imputation_helpers.

src/imputation/expansion_imputation.py
AnneONS: in imputation_main we set the default and 'R' for clear

AnneONS: and in the `apply_tmi` func in tmi_imputation sets the TMI value
                imp_class_df["imp_marker"] = "TMI"

src/imputation/expansion_imputation.py
AnneONS: This bit is the bug fix- the other question from Ana is something else again.

AnneONS:
Thanks for your review, let's get it merged! I'll hopefully also be able to merge in James' branch this morning

PR #132, opened 2023-11-17
522 fte short forms

AnneONS:
## Pull Request submission

Update apportionment so if there is  a zero in the column 202 groupby total, the apportioned column has a zero value.

Also, if the is_construction flag is true, the row is not included for imputation (but it is for apportionment and short-to-long conversion)

#### Closes or fixes

Closes #521 and #522

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/imputation/apportionment.py
```diff
@@ -43,7 +46,7 @@ def update_column(df: pd.DataFrame, col: str) -> pd.Series:
         pd.Series: A single column dataframe with the values in instance 0
         copied to other instances for the same reference.
     """
-    updated_col = df.groupby("reference")[col].transform(sum)
+    updated_col = df.groupby("reference")[col].transform(max)
```
dpurches, line 49: Do we need to add a .`order_by` here to make sure it's always instance 0? Just thinking if (for some reason) the instances were in a different order.

src/imputation/imputation_main.py
dpurches, line 1: I think the logic here makes sense - given it's outstanding whether or not constructed values should flow through estimation as well, maybe this should happen in `pipeline.py`?

src/imputation/apportionment.py
```diff
@@ -62,18 +65,21 @@ def calc_fte_column(
...
+            round_val
+        )
+        df.loc[(df["200"] == "C") & (df["tot_202_CD"] == 0), new_col] = 0
+
+        # create new apportionment column for the defence cases
+        df.loc[(df["200"] == "D") & (df["tot_202_CD"] > 0), new_col] = round(
+            update_column(df, old_cols[1]) * df["202"] / df["tot_202_CD"],
+            round_val
+        )
+        df.loc[(df["200"] == "D") & (df["tot_202_CD"] == 0), new_col] = 0
```
dpurches, line 82: I like this, much more readable 😄

dpurches (approved):
Looks good! Just a couple of questions

src/imputation/apportionment.py
AnneONS: Doesn't max take the maximum value in the group? If so, we're fine. The point is that only instance 0 has an entry, all other instances are null. I had "sum" there before, but Tom pointed out it wasn't obvious what's going on.

src/imputation/imputation_main.py
AnneONS: I'm not sure yet exactly where this should end up, I've put it after short form conversion in case short forms are "constructed" and after apportionment, assuming we'll want to apportion the constructed values.

About whether constructed values go through estimation: good point: only short form PRN included in estimation, but we should find out what to do if these include constructed values

AnneONS (commented):
Added some responses, what do you think?

src/imputation/apportionment.py
dpurches: Ah this makes sense, if it will always be the case that instance 0 is the only entry!

PR #146, opened 2023-11-27
RDRP-519: Numeric PG for SAS extract

JenCheshire:
- `check_pg_mapper.py` created to check for any inconsistencies across the PG mappers.  I have kept it in for the peer review, but it can be removed before merging.

- Long forms with numeric PG: new column "pg_numeric" created with values at staging. Column 201 is then converted to alpha, and the column name remains 201.

- The `sic_to_pg_mapper()` now takes the formtype as list.  The default is set to short forms (`0006`), but we need to be able to apply the function to imputed long forms at output

- Short forms: created an additional mapper sic_pg_num.csv from the file SIC_to_PG_UTF-8.csv, use colums SIC 2007_CODE and 2016 > Form PG; de-duplicate on SIC and check that it's a many-to-one
Stage and validate sic_pg_num.csv, use validation.validate_many_to_one

- In GB SAS output, apply the sic_pg_num to get the numerical PG for short forms and imputed long forms, and put it in pg_numeric column.   The non-imputed long form codes are kept in pg_numeric.  Also applied to the TAU output which was also using the mock mapper.

- Change the gb sas and tau schema so that the selgrpno is taken from pg_numeric column. 

Closes RDRP-519

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


check_pg_mappers.py
zorge69, line 1: Good to include this checking script. I would add de-duplication on the "many" SICside before the comparison. Still, it's not essential for live production, so happy to pass.

src/imputation/tmi_imputation.py
zorge69, line 1: I am not sure - why does it have to be changed to a list? How is it related to PG?

src/outputs/outputs_main.py
```diff
@@ -28,9 +28,8 @@ def run_outputs(
     cora_mapper: pd.DataFrame,
     postcode_mapper: pd.DataFrame,
     itl_mapper: pd.DataFrame,
-    pg_alpha_num: pd.DataFrame,
     pg_num_alpha: pd.DataFrame,
-    sic_pg_alpha: pd.DataFrame,
```
zorge69, line 33: sic_pg_alpha is still needed. We need to for short forms that are not imputed.

src/imputation/tmi_imputation.py
JenCheshire: Thanks @zorge69 I though a list might be easiest way to specify which form/s the `sic_to_pg_mapper()` should be run on.  It is used a few times in the pipeline, but on different forms, so I’ve used .isin() in the function.  e.g. `run_pg_conversion()` in staging it is only used on short forms, here in tmi_imputation it is only used on long forms, and in the outputs, it is run on both.

src/outputs/outputs_main.py
JenCheshire: I think that's being run in staging?

zorge69 (approved):
The code is looking good, but I was not able to run it because of my git issue

PR #171, opened 2023-12-28
583 consistent output

jwestw:
## Pull Request submission

The main problem I found in this area was that some of the companies which have "604" == "No" (meaning they shouldn't have R&D expenditure) also have a non-zero/positive value for "211". This is a contradiction. 

- To check the data for this problem on the way in, I created a a validator function in `validation.py` called `flag_no_rand_spenders` which is run in `staging_main.py`
- This validation function is again run later in `form_output_prep.py` to check the data for this problem at the output stage
- A [new bug ticket](https://jira.ons.gov.uk/browse/RDRP-635) was raised to check the data for this problem as it goes through the pipeline. 
- A filter was created to make sure that the records being excluded from outputs are BOTH "No" to "604" AND >0 for "211" , this is applied in `form_output_prep.py` and applied to all outputs, which makes their totals equal (at least when running on network). 

**Other changes:**
- Changed the datatype in the toml to "datetime64ns" because I had a few problems with this. I can't remember the exact problem but the pipeline would crash. Also changed the line 444 in the `validation.py` to cope with this problem. It's possible that I only needed to change one or the other to fix this bug. 
- Changed the column from "manual_trimming" to "manual_trim" which should be in common with other branches

*Insert any instructions to help the reviewer, e.g. "install new requirements from `environment.yml`"*

#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/staging/staging_main.py
```diff
@@ -365,6 +365,9 @@ def run_staging(
                 secondary_full_responses,
             )
 
+    # Flag invalid records
+    val.flag_no_rand_spenders(full_responses, "raise")
+
```
AnneONS, line 370: Do you think it's a good idea to keep this in here? Probably it is.

src/outputs/gb_sas.py
```diff
@@ -41,20 +41,17 @@ def output_gb_sas(
     paths = config[f"{NETWORK_OR_HDFS}_paths"]
     output_path = paths["output_path"]
 
-    # Filter out records that answer "no R&D"
-    df = df.copy().loc[~(df["604"] == "No")]
-
     # Filter regions for GB only
-    df = df.copy().loc[df["region"].isin(regions()["GB"])]
+    df1 = df.copy().loc[df["region"].isin(regions()["GB"])]
```
AnneONS, line 45: Is there a reason for having df1 rather than df?

src/outputs/form_output_prep.py
```diff
@@ -34,6 +35,13 @@ def form_output_prep(
         estimated_df["imp_marker"] == "R"
     )
 
+    # Deal with "No" in 604, also eliminating spenders
+    flag_no_rand_spenders(estimated_df, "warn")
+    no_rnd_spenders_filter = ~(
+        (estimated_df["604"] == "No") & (estimated_df["211"] > 0)
+    )
+    estimated_df = estimated_df.copy().loc[no_rnd_spenders_filter]
```
AnneONS, line 43: I think it might make more sense to put this in after imputation. It's only in imputation that it's possible to get these contradictory records, since we eliminated them in staging. If we leave it here, we'll have a difference between estimation_qa outputs and final outputs.

AnneONS: I think we've got 2  choices here: either we decide that if it's "check needed" and "604 = No", then we re-set the imputed values to zero (or better, don't impute them).
Alternatively, we should probably change "604" to "Yes" if it's TMI, like we do for MoR. I suggest we do this second thing for now so it will be easy to see what remains of the inconsistency

AnneONS (commented):
I think we're nearly ready, just got a question about how to treat non-zero imputed "604=No"

src/staging/validation.py
```diff
@@ -634,3 +643,30 @@ def validate_cora_df(df: pd.DataFrame) -> pd.DataFrame:
...
+    df (pandas.DataFrame): The input DataFrame.
+
+    Returns:
+        None
+    """
+    invalid_records = df.loc[(df["604"] == "No") & (df["211"] > 0)]
+
+    if not invalid_records.empty:
+        if raise_or_warn == "raise":
+            raise Exception("Some records report no R&D, but spend in 211 > 0.")
```
coatet, line 662: Would it be useful to also have a logging event here showing the total invalid spend?

coatet (commented):
One question about output for the new validation function. Happy to approve once Anne's comments are addressed.

src/imputation/imputation_main.py
```diff
@@ -99,6 +99,13 @@ def run_imputation(
     # Run TMI for long forms and short forms
     imputed_df, qa_df = tmi.run_tmi(df, mapper, config)
 
+    # After imputation, correction to ignore the "604" == "No" in any records with
+    # Status "check needed"
+    chk_mask = df["status"].str.contains("Check needed")
+    imputation_mask = df["imp_marker"].isin(["TMI", "CF", "MoR"])
+    # Changing all records that meet the criteria to "604" == "Yes"
+    df.loc[(chk_mask & imputation_mask), "604"] = "Yes"
+
```
AnneONS, line 108: That should do the trick!

AnneONS (commented):
this fix looks good!

AnneONS (approved):
checked this code gives the correct results, happy to merge

src/outputs/gb_sas.py
jwestw: I have changed this back now.

PR #174, opened 2024-01-03
RDRP-632: Update NI staging  and construction for new input schema

JenCheshire:
- New schema for NI full responses, replacing the NI indicative and responses schema files
- Update to NI construction schema to include new columns
- Remove redundant schemas/files
- To note: the GBSAS and NISAS schemas still differ slightly (turnover) - have kept as separate files for now

- Restructure staging to only read in one file, and remove the merge
- Add reg_code to NI data (currently set to N92000002, and entered separately in two places in outputs - **should this be kept in the config?**)

Closes RDRP-632 & RDRP-633


src/northern_ireland/ni_staging.py
```diff
@@ -100,14 +55,12 @@ def run_ni_staging(
...
-    then merged on reference (there are no "instance" entries in the NI data,
-    only one entry per reference.) The resulting dataframe can optionally be
-    output for qa purposes. The staged NI data is passed back to the pipeline,
-    but is not used except to be joined to the GB data for the outputs module.
+    A CSV file is read in as dataframe. Validation is performed to check
+    column names and cast to the required datatypes, based on toml schema. The
+    columns are renamed, again based on the toml schema. The resulting
+    dataframe can optionally be output for qa purposes.
+    The staged NI data is passed back to the pipeline, but is not used
+    except for the outputs module.
```
AnneONS: Perhaps we should mention NI construction too:

The staged NI data is passed back to the pipeline for optional NI construction.
Apart from construction, the NI data is not used for other pipeline modules 
before the outputs module.

AnneONS (commented):
minor suggestion for docstrings, I just need to test the NI construction and I think we're close to merging.

AnneONS (commented):
The employment emp_xx cols are missing from NI sas for the constructed records...

We need a correction to NI construction: we need to include the 4xx and 5xx columns. This is because they are needed for NI apportionment. It might be worth checking in the construction code why existing 4xx data gets replaced with nulls currently.

Or can you think of a better way to do this?

AnneONS (approved):
Checked the results and all works as expected

PR #186, opened 2024-01-17
RDRP-667: Construction bug fixes

JenCheshire:
**short_to_long - already in v0.6.2 and tested by Dan**
If there were no short_to_long constructions the flag column would be dropped causing an error when it got to short_to_long prep.
An if statement has been added so it only prepares for short_to_long if there are short forms to be converted to long.
e.g. 49900102021 - should be showing three rows, formtype 6 after construction

**short forms with no instance**
short forms with status 'form sent out' do not have an instance, so are not being matched to the relevant construction record as they have an instance 0 in the construction file.
The indexing requires there to be an instance in both construction and in the snapshot, so a temp instance is created and then removed from those records in the snapshot to match construction file.
e.g. 49900100508 - change to XX11 1AA from WV13 1RD


jwestw (approved):
The code all runs now and has been checked over. Happy to approve and merge.

PR #187, opened 2024-01-17
RDRP-669: statusencoded/CORA bug fixes

JenCheshire:
Fixed the bugs that had stopped the develop pipeline from working, and CORA statuses not being mapped correctly.
Changed the statusencoded schema back to category (which stops the error in TMI), and changed the CORA dictionary to strings (which allows statuses to be mapped correctly and appear in outputs).

Also small fix to the datetime64[ns] in schemas to allow it to run (needs square brackets).

This has been tested with and without the feather.


zorge69 (approved):
The pipeline runs. 
The change of datatype for status encoded to category is correct.

PR #191, opened 2024-01-23
Update to construction schema

JenCheshire:
Resetting the instance schema to Int64 in construction - it was changed to float in the last version, however this was causing problems for the index/.update as snapshot schema was Int64 so was not matching.
Also pushed for review on DAP

24/01: added an if statement to the formtype convert function.  Dan tried a postcode only construction file, and because formtype col dropped, it caused an error.  This is running correctly in my profile space on DAP.


AnneONS (approved):
All good

PR #192, opened 2024-01-23
649 imputation class threshold for MoR

AnneONS:
## Pull Request submission

In mean of ratios, created a configurable threshold value for the size of an imputation class, based on the number of non-zero values in col 211.

If the imputation class is below the threshold, a fall-back to the carry forwards method is used instead.

Note for reviewer: An extra column, cf_group_size has been added to the links QA output csv, this shows the size of the group, and if it's below the threshold chosen (I suggest 3 for the network, so you have examples below and above) then the link should be 1.  Please could you check that the sizes in that column are correct? The number of positive values in the column 211 for that imputation class. (I have checked and it seems ok)

Another fix: previously we filled nulls with zeros for all carry forward rows, this time we only fill nulls with zeros where column 211 is not null. This means that "post code only" and instance 0 rows will remain null for the key variables.
 

#### Closes or fixes

Closes #649

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)

### DIFF
--- a/src/developer_config.yaml
+++ b/src/developer_config.yaml
@@ -23,7 +23,7 @@ global:
   # Output settings
   output_full_responses: False
   output_ni_full_responses: False
-  output_imputation_qa: False
+  output_imputation_qa: True
   output_auto_outliers: False
   output_outlier_qa : False
   output_estimation_qa: False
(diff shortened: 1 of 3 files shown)


JenCheshire:
@AnneONS `cf_group_size` really helpful in links_qa. An empty `cf_group_size` column is in the imputation_qa - can that be taken out to avoid confusion?

JenCheshire (approved):
Looks good, just the col suggestion .  Can see the difference between the df's when threshoold is set to 3, compared to 0.   Did try a higher threshold, but the imp_classes are limited - happy to look at that in DAP in the next version

PR #194, opened 2024-01-26
RDRP-667: short to long construction bug fix

JenCheshire:
Have updated the `prep_short_to_long()` function, so that it reads the number of instances required, and creates the correct number by copying the instance 0 (updating `instance` for each instance). 
Added an example with 5 instances in the test construction file.

A few formatting changes from the pre-commit too


AnneONS (approved):
Good solution! I've checked the code and the outputs, it's working as it should.

PR #197, opened 2024-02-01
RDRP-681:  NI form_status in TAU bug fix

JenCheshire:
The `form_status` is being given to NI data in `form_output_prep.py`, so it can be passed directly to NI SAS.  However, in the TAU output, the `form_status` for GB data is being mapped from `statusencoded` - because NI data doesn't have a `statusencoded` the `form_status` was being overwritten with a null

Have changed the code in `create_cora_status_col()` so that mapping only occurs in the rows where `form_status` is null, which retains the NI form_status.  Because the function is also used by GB SAS, which will not have the NI data and so no `form_status` column, it checks if the column exists and adds it before the mapping function to allow it to work.

### DIFF
--- a/src/outputs/map_output_cols.py
+++ b/src/outputs/map_output_cols.py
@@ -137,8 +137,14 @@ def create_cora_status_col(df, main_col="statusencoded"):
 
     cora_dict = dict(zip(status_before, status_after))
 
-    # Create a new column by mapping values from main_col using the cora_dict
-    df["form_status"] = df[main_col].map(cora_dict)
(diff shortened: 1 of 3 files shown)


src/outputs/map_output_cols.py
zorge69, line 1: I agree with the logic. The code runs and produces ni_sas and tau outputs
NI records in NI SAS have reg_code N92000002
All records with reg_code N92000002 in TAU file have rec_mkr == 600.

zorge69 (approved):

PR #199, opened 2024-02-02
RDRP-687: Cell 817 mapping fix

JenCheshire:
We changed all formtypes to string in the schemas, so the condition was no longer matching for the filter in `update_ref_list()`.  This meant there were no ref's to map.  Have changed the formtype to a string.
Because it was such a small change, I have also tested in DAP and all the examples are showing 817.


zorge69 (approved):
Good fix!
After the delivery, we would need to apply a holistic approach to all codes.
The pipeline runs end to end in Windows.
Checked staged_BERD_full_responses_2024-02-02_v105.csv, and the selected refs come out with corrected cellnumber.

PR #200, opened 2024-02-02
RDRP-688: No R&D in GB SAS

JenCheshire:
There were 'No R&D' (`604` = No) in the GB SAS output.  There is a filter in the `form_output_prep()`, but it is only taking out records with `211` > 0.  

There is now an explicit filter at the beginning of GB SAS (so we don't lose these records elsewhere) that takes out `604` = No. This is already present in the TAU output.

### DIFF
--- a/src/outputs/gb_sas.py
+++ b/src/outputs/gb_sas.py
@@ -35,9 +35,6 @@ def output_gb_sas(
     paths = config[f"{NETWORK_OR_HDFS}_paths"]
     output_path = paths["output_path"]
 
-    # Filter out records that answer "no R&D"
-    df = df.copy().loc[~(df["604"] == "No")]
-
     # Filter regions for GB only
     df1 = df.copy().loc[df["region"].isin(regions()["GB"])]
 
--- a/src/outputs/gb_sas.py
+++ b/src/outputs/gb_sas.py
@@ -35,6 +35,9 @@ def output_gb_sas(
     paths = config[f"{NETWORK_OR_HDFS}_paths"]
     output_path = paths["output_path"]
 
+    # Filter out records that answer "no R&D"
+    df = df.copy().loc[~(df["604"] == "No")]
+
     # Filter regions for GB only
     df1 = df.copy().loc[df["region"].isin(regions()["GB"])]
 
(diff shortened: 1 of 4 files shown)


src/outputs/gb_sas.py
```diff
@@ -35,6 +35,9 @@ def output_gb_sas(
     paths = config[f"{NETWORK_OR_HDFS}_paths"]
     output_path = paths["output_path"]
 
+    # Filter out records that answer "no R&D"
+    df = df.copy().loc[~(df["604"] == "No")]
+
```
AnneONS: can I just check, could or should we put this in output_main instead?

JenCheshire: Good question @AnneONS .  Is this relevant to all outputs, inc. short form and long form outputs?  If so, it could go in `form_output_prep()`.  If it should be included for SAS, TAU and intram outputs, it could go in `outputs_main()`.  @zorge69 please could you confirm?

JenCheshire: @AnneONS That change has been made

src/outputs/outputs_main.py
```diff
@@ -100,6 +100,10 @@ def run_outputs(
         )
         OutputMainLogger.info("Finished long form output.")
 
+    # Filter out records that answer "no R&D" for all subsequent outputs
+    tau_outputs_df = tau_outputs_df.copy().loc[~(tau_outputs_df["604"] == "No")]
+    outputs_df = outputs_df.copy().loc[~(outputs_df["604"] == "No")]
+
```
AnneONS, line 106: @JenCheshire  I just updated this from ~(df[   which needed correcting, can you just check I've done the right thing? Then I can approve.

JenCheshire: @AnneONS Yes, that's right - thank you for spotting that!

src/outputs/outputs_helpers.py
```diff
@@ -102,6 +102,7 @@ def create_period_year(df: pd.DataFrame) -> pd.DataFrame:
     """
 
     # Extracted the year from period and crated new columns 'period_year'
+    df = df.copy()
```
zorge69, line 105: I don't understand why we need it.

zorge69 (commented):
Looks good to me, happy to merge

AnneONS (approved):
checked output as well as code, it's ready

PR #201, opened 2024-02-02
RDR-689: Funding totals not equaling

JenCheshire:
Totals were not equaling the sum of the breakdown columns in DAP.  This was due to nulls in the breakdown columns, causing the sum to equal null,even where some had values.  Have added .fillna(0) to the breakdown columns in the sum.
Tested in my profile of DAP, and the total cols are now showing (example, in GB SAS v306 & v307).

These columns are being created in GB SAS, NI SAS and TAU, and I have just added the .fillna() to each - I didn't move into its own function at this point, because I didn't want to make too many changes - let me know if you think I should do this now though

### DIFF
--- a/src/developer_config.yaml
+++ b/src/developer_config.yaml
@@ -26,13 +26,13 @@ global:
   output_imputation_qa: False
   output_auto_outliers: False
   output_outlier_qa : False
-  output_estimation_qa: True
-  output_apportionment_qa: True
+  output_estimation_qa: False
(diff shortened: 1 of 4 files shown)


zorge69 (approved):
The conversion Nulls to 0 should work. I liked the summation along 1 axis, it makes the code more elegant.

The code runs in Windows. It creates numerical sums in gb_sas output. We need to build it and re-test it in DAP.

PR #206, opened 2024-02-09
RDRP-671: Period year for SAS & TAU outputs

JenCheshire:
Moved the `create_period_year()` function to `form_output_prep()`, so it can be pulled straight into all relevant outputs.  Adjusted schema's so all relevant outputs pull the period year rather than period.  
Moved the unit test into 'test_output_helpers()` so the it mirrors the code.

### DIFF
--- a/config/output_schemas/gb_sas_schema.toml
+++ b/config/output_schemas/gb_sas_schema.toml
@@ -3,7 +3,7 @@ old_name = "reference"
 Deduced_Data_Type = "Int64"
 
 [period]
-old_name = "period"
+old_name = "period_year"
 Deduced_Data_Type = "Int64"
 
 [entref]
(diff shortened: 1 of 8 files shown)


config/output_schemas/gb_sas_schema.toml
```diff
@@ -3,7 +3,7 @@ old_name = "reference"
 Deduced_Data_Type = "Int64"
 
 [period]
-old_name = "period"
+old_name = "period_year"
```
zorge69, line 6: It's nice that we can choose what column goes into output.
The risk is, the toml files are not part of the package, so they need to be updated manually of the v_0.7.0 of the runner.

zorge69 (approved):
Pipeline runs without errors. Checked NI SAS, GB SAS and Tau outputs, and the period is just the year, four digits. NI period is 2021 but this is OK, on DAP it should be 2022.

PR #215, opened 2024-02-22
RDRP-651: Unit tests for calculate_weights and apply_weights

JenCheshire:
Four tests for `calc_lower_n()`:
testing calc_lower_n where duplicate refs
testing calc_lower_n where missing col
testing calc_lower_n where nan in reference

Five tests for `calculate_weighting_factor()`:
testing calculate_weighting_factor where missing outlier col
testing calculate_weighting_factor filter (rtol=0.01)
testing calculate_weighting_factor 709 to numeric with no missing vals
testing calculate_weighting_factor 709 to numeric with missing vals
testing calculate_weighting_factor with missing vals

One test for `outlier_weights()`:
test that all appropriate rows are given an a_weight = 1.0

One test for `test_apply_weights()`:
test that estimation cols are created, correctly calculated and rounded (rtol=0.05).  Includes a np.nan and 0.  Includes a test col_list.


tests/test_estimation/test_apply_weights.py
```diff
@@ -0,0 +1,143 @@
...
+    ]
+
+    exp_output_df = pd.DataFrame(data=data, columns=exp_output_columns)
+
+    result_df, cols_list = appw.apply_weights(input_df, config, 4)
+
+    input_columns.remove("reference"); input_columns.remove("a_weight")
+
+    assert_frame_equal(result_df, exp_output_df, check_like=True, check_exact=False, rtol=0.05)
+    assert cols_list.sort() == input_columns.sort()
```
cbrowning04: Minor thing but do we want to add a message to this assertion? This will help to identify the issue if this test is ever broken.
(Optional implementation)

tests/test_estimation/test_calculate_weights.py
```diff
@@ -0,0 +1,454 @@
...
+        [2, 'C'],
+        [4, 'D'],
+        [1, 'E'],
+    ]
+
+    input_df = pd.DataFrame(data=data, columns=input_cols)
+
+
+    with pytest.raises(ValueError):
+        calw.calc_lower_n(input_df)
```
cbrowning04: Can this be amended to account for both cases? Along with message checking for pytest.raises.
Here is an (OPTINONAL) implementation:
```python
@pytest.parameterize(
    "cols, msg, exp_col",
    (
        [["ref", "709"], f"'reference' or 709 missing.", "709"],
        [["reference", "706"], f"'reference' or 707 missing.", "707"]
    )
)
def test_calc_lower_n_missing_col(cols, msg, exp_col):
    """Test for calc_lower_n with missing col"""
    data = [
        [1, 'A'],
        [2, 'B'],
        [2, 'C'],
        [4, 'D'],
        [1, 'E'],
    ]
    input_df = pd.DataFrame(data=data, columns=cols)

    with pytest.raises(ValueError, msg=msg):
        calw.calc_lower_n(input_df, exp_col)
```
NOTE: The above code is untested.

tests/test_estimation/test_calculate_weights.py
```diff
@@ -0,0 +1,454 @@
...
+    actual_result = calw.calc_lower_n(input_df)
+
+    # Defined expected result
+    expected_result = 3
+
+    assert actual_result == expected_result
+
+
+def test_calc_lower_n_nan_col():
+    """Test for calc_lower_n with nan in reference."""
```
cbrowning04: Docstring copied from above here

tests/test_estimation/test_calculate_weights.py
```diff
@@ -0,0 +1,454 @@
...
+
+    input_df = pd.DataFrame(data=data, columns=input_cols)
+
+    # Call calc_lower_n function
+    actual_result = calw.calc_lower_n(input_df)
+
+    # Defined expected result
+    expected_result = 3
+
+    assert actual_result == expected_result
```
cbrowning04: As the function is testing the count of unique references, this test is less relevant when we are checking for nan's in the exp_col. Maybe a test for `does_not_raise()` would be more useful here?

For clarity, the below can be used to test:
```python
from contextlib import nullcontext as does_not_raise
with does_not_raise():
    # do something
```

tests/test_estimation/test_calculate_weights.py
```diff
@@ -0,0 +1,454 @@
...
+        [1, 'A'],
+        [2, 'B'],
+        [2, 'C'],
+        [4, 'D'],
+        [1, 'E'],
+    ]
+
+    input_df = pd.DataFrame(data=data, columns=input_cols)
+
+    with pytest.raises(ValueError):
```
cbrowning04: message in pytest.raises

tests/test_estimation/test_calculate_weights.py
```diff
@@ -0,0 +1,454 @@
...
+        "formtype",
+        "instance",
+        "cellnumber",
+        "outlier",
+    ]
+
+    data = [
+        [1, '12', "P", "210", "0006", 0, 1, False],
+        [2, '14', "P", "211", "0006", 0, 2, False],
+        [2, 16, "P", "210", "0006", 0, 2, False],
```
cbrowning04: Shouldn't there be a missing value here in the '709' column?

tests/test_estimation/test_calculate_weights.py
```diff
@@ -0,0 +1,454 @@
...
+        "formtype",
+        "instance",
+        "cellnumber",
+        "outlier",
+    ]
+
+    data = [
+        [1, '12', "P", "210", "0006", 0, 1, False],
+        [2, np.nan, "P", "211", "0006", 0, 2, False],
+        [2, 16, "P", "210", "0006", 0, 2, False],
```
cbrowning04: It appears that this and the above function names should be corrected since this has an np.nan while the above doesn't.
Happy to be corrected on this since I might be missing something

tests/test_estimation/test_calculate_weights.py
```diff
@@ -0,0 +1,454 @@
...
+        [1, True, 1.0],
+        [2, False, None],
+        [2, True, 1.0],
+        [4, True, 1.0],
+        [1, False, None],
+    ]
+
+    expected_df = pd.DataFrame(data=data, columns=expected_cols)
+
+    assert_frame_equal(result_df, expected_df)
```
cbrowning04: if you wanted to save lines here, you could just assert the series for a_weight. This doesn't make a difference to the actual test however so just optional :).
Also, it might not be useful in the context of the functions use, however it appears that there are no defences for cases where the outlier column isn't present

cbrowning04 (changes requested):
Hi Jen great stuff. All tests do as expected and cover the relevant areas.

Review Notes:
- Possibly create classes for each of the functions being tested to group the individual tests
- Assertion messages are useful for future debugging
- Add error message checking to ensure the correct error is being raised (not only the correct type)
- Other small areas covered in the review comments

tests/test_estimation/test_calculate_weights.py
JenCheshire: Good spot!  I've mixed the names

cbrowning04:
Hi @JenCheshire . Great changes. Will get this merged today

cbrowning04 (approved):

PR #223, opened 2024-03-12
RDRP-750 and 751 tests for create_sites_df and count_duplicate_sites

cbrowning04:
## Pull Request submission

This PR addresses  RDRP-750 and RDRP-751. This includes:
- Tests for `site_apportionment::site_apportionment.create_sites_df`.
- Tests for `site_apportionment::site_apportionment.count_duplicate_sites`.

Additional changes in this PR includes:
- Updating inconsistent docstrings throughout `site_apportionment::site_apportionment`.
- Re-ordering some imports to comply with PEP-8.

#### Code

- [x] **Clean Code**

#### Data
- [x] All data needed to run this script is available in Dev/Test
- [x] All data is excluded from this pull request
- [x] Secrets checker pre-commit passes

#### Testing
- [x] **Unit tests** Unit tests have been created and are passing _or a new ticket to create tests has been created_

---

# Peer Review Section

---

### Review comments

  works correctly?)

### DIFF
--- a/tests/test_site_apportionment/test_site_apportionment.py
+++ b/tests/test_site_apportionment/test_site_apportionment.py
@@ -670,10 +670,12 @@ class TestCreateSitesDf(object):
         ]
         input_data = [
             [1, 0, "0006", "RH12 1XL", 100.0, np.nan, "Clear", "R", "RH12 1XL", "202101"],
-            [1, 1, "0006", "RH12 1XL", 100.0, np.nan, "Clear", "R", "RH12 1XL", "202101"],
-            [1, 2, "0006", "NP44 2NZ", 100.0, np.nan, "Clear", "R", "NP44 2NZ", "202101"],
+            [1, 1, "0006", "RH12 1XL", 125.0, np.nan, "Clear", "R", "RH12 1XL", "202101"],
+            [1, 2, "0006", "RH12 1XL", np.nan, np.nan, "Clear", "R", "RH12 1XL", "202101"], # Nan 602 - Ensure conv to 0
+            [1, 2, "0006", "RH12 1XZ", 100.0, np.nan, "Clear", "R", "RH12 1XZ", "202101"], # different postcode
             [2, 0, "0001", "NP44 2NZ", np.nan, 2.0, "Clear", "R", "NP44 2NZ", "202102"],
-            [3, 0, "0001", np.nan, np.nan, 1.0, "Check needed", "TMI", "NP30 7ZZ", "202102"],
+            [2, 1, "0001", "NP44 2NZ", 50.0, 2.0, "Clear", "R", "NP44 2NZ", "202102"],
+            [3, 0, "0001", np.nan, np.nan, 1.0, "Check needed", "TMI", "NP30 7ZZ", "202102"], # NaN 601 - Ensure dropped
         ]
         input_df = pandasDF(data=input_data, columns=input_cols)
 
@@ -711,13 +713,13 @@ class TestCreateSitesDf(object):
                             "602",
                             "postcodes_harmonised"]
         assert sorted(expected_columns) == sorted(output.columns.values)
-        assert np.array_equal(output["602"], [100.0, 200.0, 0.0]), (
+        assert np.array_equal(output["602"], [225.0, 100.0, 50.0]), (
             "Column 602 (percent_col) has unexpected values."
         )
-        assert np.array_equal(output["instance"], [2, 0, 0]), (
+        assert np.array_equal(output["instance"], [0, 2, 0]), (
             "Column 602 (percent_col) has unexpected values."
         )
-        expected_postcodes = ["NP44 2NZ", "RH12 1XL"]
+        expected_postcodes = ["NP44 2NZ", "RH12 1XL", "RH12 1XZ"]
         assert np.array_equal(
             sorted(output["601"].unique()),
             sorted(expected_postcodes)
--- a/tests/test_site_apportionment/test_site_apportionment.py
+++ b/tests/test_site_apportionment/test_site_apportionment.py
@@ -641,7 +641,14 @@ class TestCreateSitesDf(object):
 
     @pytest.fixture(scope="function")
     def create_sites_df_input(self):
-        """Input df for create_sites_df tests."""
+        """
+        Input df for create_sites_df tests.
+        
+        This input data consists of:
+        1. Columns that will be dropped
+        2. NaN's in used columns (601, 602)
+        3. Combinations of Postcodes/Periods/References to create multiple groups
+        """
         input_cols = [
             "reference",
             "instance",
@@ -654,12 +661,11 @@ class TestCreateSitesDf(object):
             "postcodes_harmonised",
             "period"
         ]
-
         input_data = [
-            [1, 0, "0006", "RH12 1XL", 100.0, np.nan, "Clear", "R", "NP10 5XX", "202101"],
-            [1, 1, "0006", "RH12 1XL", 100.0, np.nan, "Clear", "R", "NP10 5XX", "202101"],
-            [1, 2, "0006", "NP44 2NZ", 100.0, np.nan, "Clear", "R", "NP10 5XX", "202101"],
-            [2, 0, "0001", "NP44 2NZ", np.nan, 2.0, "Clear", "R", "NP20 6YY", "202102"],
+            [1, 0, "0006", "RH12 1XL", 100.0, np.nan, "Clear", "R", "RH12 1XL", "202101"],
+            [1, 1, "0006", "RH12 1XL", 100.0, np.nan, "Clear", "R", "RH12 1XL", "202101"],
+            [1, 2, "0006", "NP44 2NZ", 100.0, np.nan, "Clear", "R", "NP44 2NZ", "202101"],
+            [2, 0, "0001", "NP44 2NZ", np.nan, 2.0, "Clear", "R", "NP44 2NZ", "202102"],
             [3, 0, "0001", np.nan, np.nan, 1.0, "Check needed", "TMI", "NP30 7ZZ", "202102"],
         ]
         input_df = pandasDF(data=input_data, columns=input_cols)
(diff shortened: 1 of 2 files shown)


tests/test_site_apportionment/test_site_apportionment.py
```diff
@@ -1,20 +1,29 @@
...
-    sort_rows_order_cols
+    sort_rows_order_cols,
+    create_sites_df,
+    count_duplicate_sites
 )
 
+# Define easier pandas usages
+pandasDF = pd.DataFrame
+assert_frame_equal = pd._testing.assert_frame_equal
+assert_series_equal = pd._testing.assert_series_equal
```
AnneONS, line 27: Nice! Thank you.

tests/test_site_apportionment/test_site_apportionment.py
```diff
@@ -634,3 +643,112 @@ def test_sort_rows_missing_cols(self, caplog):
...
+        "status",
+        "imp_marker",
+        "postcodes_harmonised",
+        "period"
+    ]
+    input_data = [
+        [1, 0, "0006", "RH12 1XL", 100.0, np.nan, "Clear", "R", "RH12 1XL", "202101"],
+        [1, 1, "0006", "RH12 1XL", 125.0, np.nan, "Clear", "R", "RH12 1XL", "202101"],
+        [1, 2, "0006", "RH12 1XL", np.nan, np.nan, "Clear", "R", "RH12 1XL", "202101"], # Nan 602 - Ensure conv to 0
+        [1, 2, "0006", "RH12 1XZ", 100.0, np.nan, "Clear", "R", "RH12 1XZ", "202101"], # different postcode
```
AnneONS: In this case, the instance would be 3, as before apportionment is applied, the [reference, instance] combination is unique.  I'm not sure of your comment here, as the postcode isn't different on this line?

tests/test_site_apportionment/test_site_apportionment.py
```diff
@@ -634,3 +643,112 @@ def test_sort_rows_missing_cols(self, caplog):
...
+        self.groupby_cols = ["reference", "period"]
+        self.site_cols = ["instance", "601", "602", "postcodes_harmonised"]
+    
+    def test_create_sites_df_raises(self, sites_df_input):
+        """Tests for create_sites_df when invalid data is passed."""
+        # test when a required column is missing
+        no_period = sites_df_input.copy()
+        no_period.drop("period", axis=1, inplace=True)
+        with pytest.raises(KeyError, match=r".*period.*not in index"):
+            create_sites_df(no_period, self.groupby_cols, self.site_cols)
```
AnneONS: I wouldn't have thought to test this. Isn't it just an attribute of python that this behaviour will occur?

tests/test_site_apportionment/test_site_apportionment.py
```diff
@@ -634,3 +643,112 @@ def test_sort_rows_missing_cols(self, caplog):
...
+        # cols passed not in list[str] format
+        with pytest.raises(TypeError, match="unsupported operand type.*"):
+            create_sites_df(sites_df_input, 3, self.site_cols)
```
AnneONS: This is good, we should use this more often.

tests/test_site_apportionment/test_site_apportionment.py
```diff
@@ -634,3 +643,112 @@ def test_sort_rows_missing_cols(self, caplog):
...
+class TestCreateSitesDf(object):
+    """Tests for create_sites_df."""
+    
+    @pytest.fixture(autouse=True)
+    def set_attrs(self):
+        """Set class attributes that are passed as parameters."""
+        self.groupby_cols = ["reference", "period"]
+        self.site_cols = ["instance", "601", "602", "postcodes_harmonised"]
+    
+    def test_create_sites_df_raises(self, sites_df_input):
```
AnneONS: This function includes some really useful utils that we should be using in our pipeline, thank you for introducing them!

however, do you think that including error catching for unit tests might be a bit over kill if they're fairly short and not too complex? If we added something like this to all our unit tests it would make them longer, and possibly that would mean they are not so readable? What do you think? Will it significantly help debugging unit tests?

tests/test_site_apportionment/test_site_apportionment.py
```diff
@@ -634,3 +643,112 @@ def test_sort_rows_missing_cols(self, caplog):
...
+            "Column 602 (percent_col) has unexpected values."
+        )
+        assert np.array_equal(output["instance"], [0, 2, 0]), (
+            "Column 602 (percent_col) has unexpected values."
+        )
+        expected_postcodes = ["NP44 2NZ", "RH12 1XL", "RH12 1XZ"]
+        assert np.array_equal(
+            sorted(output["601"].unique()),
+            sorted(expected_postcodes)
+            ), "Postcodes not as expected"
```
AnneONS: This approach is thorough. However, on this project our style has been to create the expected_output dataframe. I think this is in general a bit easier to read, as you can see at a glance what the shape and content of the expected output. This means the unit test also acts as an illustration of the action of the function. Would you be happy to update to this style?

tests/test_site_apportionment/test_site_apportionment.py
```diff
@@ -634,3 +643,112 @@ def test_sort_rows_missing_cols(self, caplog):
...
+    """Tests for count_duplicate_sites."""
+
+    def test_count_duplicate_sites_raises(self, sites_df_input):
+        """Tests for raises from bad data passed to count_duplicate_sites."""
+        # passing dataframe with missing columns
+        with pytest.raises(KeyError, match=".*reference.* not in index"):
+            count_duplicate_sites(sites_df_input.drop("reference", axis=1))
+        # passing wrong type (not a df)
+        with pytest.raises(TypeError, match=".*int.* object is not subscriptable"):
+            count_duplicate_sites(0)
```
AnneONS: I like the way you've done these tests for errors. We can use this pattern again.

AnneONS (commented):
Really good work, thanks Charlie! Just a few comments, mostly concerning our style.

tests/test_site_apportionment/test_site_apportionment.py
cbrowning04: Hi Anne. Will change the instance on this. The postcodes are different - `"RH12 1XL"` and `"RH12 1XZ"`

tests/test_site_apportionment/test_site_apportionment.py
cbrowning04: Error catching is very important to have in unit tests, however these functions do not raise any errors and have no defences (you can pass anything to it and it won't be checked for any mishaps e..g, datatype, length, contents etc), therefore they might not be as useful here.

tests/test_site_apportionment/test_site_apportionment.py
cbrowning04: I could update it to that style, however I don't necessarily agree with it when you are looking to tests various parts and different inputs. Redefining input and output dfs many times over and over clutters the test file and leaves your test script as mainly just hard coded data.
In my opinion, asserting the output with corresponding failure message is sufficient, as the goal of the unit test is to ensure that the function is producing the expected output so that you know the functions work correctly.

tests/test_site_apportionment/test_site_apportionment.py
cbrowning04: Done.

tests/test_site_apportionment/test_site_apportionment.py
AnneONS: OK, so I suggest we remove it from these unit tests, and instead focus on adding defences in utils in future, as I'd prefer to keep the unit tests shorter and hence more readable.

tests/test_site_apportionment/test_site_apportionment.py
AnneONS: Hi Charlie, you make a good point, and I agree that the first aim of the unit test is to ensure each aspect of the test is covered. However, for now I think I'd prefer  us to follow what I take to be the style we use in the various RDSA projects I've worked on, which is to have hard coded input and expected output dataframes. 
I think the advantage of this is that it allows people with limited coding experience to see at a glance what the function does.

Let's take the debate on different ways of doing unit tests to an RDSA best practice chat some time :-)

AnneONS:
You've done some really fab work on these unit tests, it's so good to have you on the team. For now though, please could you make those two amendments and I'm happy to continue debate on how best to do unit testing going forwards.

tests/test_site_apportionment/test_site_apportionment.py
cbrowning04: It's important that we know that a function does not operate when key data is missing.

tests/test_site_apportionment/test_site_apportionment.py
```diff
@@ -634,3 +643,87 @@ def test_sort_rows_missing_cols(self, caplog):
...
+    ]
+    input_data = [
+        [1, 0, "0006", "RH12 1XL", 100.0, np.nan, "Clear", "R", "RH12 1XL", "202101"],
+        [1, 1, "0006", "RH12 1XL", 125.0, np.nan, "Clear", "R", "RH12 1XL", "202101"],
+        [1, 2, "0006", "RH12 1XL", np.nan, np.nan, "Clear", "R", "RH12 1XL", "202101"], # Nan 602 - Ensure conv to 0
+        [1, 3, "0006", "RH12 1XZ", 100.0, np.nan, "Clear", "R", "RH12 1XZ", "202101"], # different postcode
+        [2, 0, "0001", "NP44 2NZ", np.nan, 2.0, "Clear", "R", "NP44 2NZ", "202102"],
+        [2, 1, "0001", "NP44 2NZ", 50.0, 2.0, "Clear", "R", "NP44 2NZ", "202102"],
+        [3, 0, "0001", np.nan, np.nan, 1.0, "Check needed", "TMI", "NP30 7ZZ", "202102"], # NaN 601 - Ensure dropped
+    ]
```
AnneONS, line 679: For illustration of what the pipeline is like at this point, please don't include formtype or instance 0

AnneONS (approved):
Ready to merge, many thanks

PR #254, opened 2024-06-27
RDRP-857: create mapping module

JenCheshire:
New module called mapping and a new script src.mapping.mapping_main.py (called from pipeline.py after construction). Passes in full_responses and returns mapped_df.

Moved all functions for the staging of mappers (except detailed output mappers) from staging_main.py to new script, src.mapping.mapping_main.py

Moved validation functions (where not used in staging) called by these functions to a new script src.mappping.mapping_helpers.py

Return commented out (although imputation input changed to mapped_df in pipeline) and placeholder for actual mapping

## Pull Request submission

#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/mapping/mapping_helpers.py
```diff
@@ -0,0 +1,91 @@
...
+import pandas as pd
+import logging
+
+from src.utils.wrappers import time_logger_wrap, exception_wrap
+
+MappingLogger = logging.getLogger(__name__)
+
+
+@time_logger_wrap
+@exception_wrap
```
AnneONS: shall we take out these wrappers, do you think? I'm not sure they do much, and they make it more difficult using the debugger, as James said.

src/mapping/mapping_main.py
```diff
@@ -0,0 +1,125 @@
...
+def run_mapping(
+    full_responses,
+    config: dict,
+    check_file_exists: Callable,
+    load_json: Callable,
+    read_csv: Callable,
+    write_csv: Callable,
+    read_feather: Callable,
+    write_feather: Callable,
+    isfile: Callable,
```
AnneONS: shall we import these directly into this module, rather than passing them?

tests/test_mapping/test_mapping_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+    @pytest.fixture(scope="function")
+    def full_input_df(self):
+        """Main input data for update_ref_list tests."""
+        columns = ["reference", "instance", "formtype", "cellnumber"]
+        data = [
+            [49900001031, 0.0, 6, 674],
+            [49900001530, 0.0, 6, 805],
+            [49900001601, 0.0, 1, 117],
+            [49900001601, 1.0, 1, 117],
+            [49900003099, 0.0, 6, 41],
```
AnneONS: it's perhaps nicer to have an integer for instance. There should be some examples of tests where we've got round the problem of it creating a different type to the one we expect, for example using .astype or ignoring the types when we do the assert.

tests/test_mapping/test_mapping_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+        ]
+        df = pd.DataFrame(columns=columns, data=data)
+        df["formtype"] = df["formtype"].apply(lambda x: str(x))
```
AnneONS: does it work to do df = df.astype({"formtype":int}) or something like that?

AnneONS (commented):
Some minor initial comments, I'll have a deeper look at the tests in a mo :-)

src/mapping/mapping_main.py
JenCheshire: Ah, I think these being passed, rather than imported, because they will change depending on whether it is local v hdfs?

AnneONS: yes, I think we were talking about simplifying it to     

    network_or_hdfs = config["global"]["network_or_hdfs"]

    if network_or_hdfs == "network":
        from src.utils import local_file_mods as mods

    elif network_or_hdfs == "hdfs":
        from src.utils import hdfs_mods as mods

what do you think?

JenCheshire: Yes, you're right - we spoke about doing that in the module! Sorry, had slipped my mind

tests/test_mapping/test_mapping_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+from src.mapping.mapping_helpers import update_ref_list
+
+
+class TestUpdateRefList(object):
+    """Tests for update_ref_list."""
+
+    @pytest.fixture(scope="function")
+    def full_input_df(self):
+        """Main input data for update_ref_list tests."""
+        columns = ["reference", "instance", "formtype", "cellnumber"]
```
AnneONS: I think we should have selectiontype in here as well, so we don't have nulls in the expected dataframe at the end//

tests/test_mapping/test_mapping_helpers.py
AnneONS: I think maybe for the test we should have our formtype values as "0006" and "0001" as they are in the full responses.

tests/test_mapping/test_mapping_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+        df = pd.DataFrame(columns=columns, data=data)
+        df["formtype"] = df["formtype"].apply(lambda x: str(x))
+        return df
+
+    @pytest.fixture(scope="function")
+    def expected_output(self):
+        """Expected output for update_ref_list tests."""
+        columns = ["reference", "instance", "formtype", "cellnumber", "selectiontype"]
+        data = [
+            [49900001031, 0.0, "6", 674, np.nan],
```
AnneONS: again, I think formtype needs to be "0006" and "0001"

tests/test_mapping/test_mapping_helpers.py
JenCheshire: I've changed instance to an int - but formtype is handled as a string

AnneONS (commented):
A couple more minor comments here

AnneONS (approved):
All good! many thanks

PR #273, opened 2024-07-19
RDRP-921 output backdata for imputation

AnneONS:
## Pull Request submission

The code now outputs data to used for forward imputation in the next year.

As part of this, there's a temp script to deal with the special case of 2021 data. This creates a new version of the 2021 data which has everything needed for imputation going forwards. When we've settled on how we do forward imputation, and everything is also set up in DAP, this can be deleted.

Other minor improvements to the pipeline have been included, and the precommit yaml updated so tests are excluded from limiting line lengths.

#### Closes or fixes

Closes #921

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/outputs/outputs_main.py
```diff
@@ -25,7 +25,7 @@
 
 
 def run_outputs(  # noqa: C901
-    estimated_df: pd.DataFrame,
```
JenCheshire: Why isn't this needed?

src/site_apportionment/site_apportionment_main.py
```diff
@@ -25,14 +24,12 @@ def run_site_apportionment(
     instance 0 to all other instances.
     Same percentages are used for each product group.
 
-    When running on the local network,
```
JenCheshire, line 28: I know this isn't part of this ticket - should we record the tidying up jobs that need doing through the pipeline?  The docstring needs updating for this function.

JenCheshire (approved):
Only two minor things - looks really good, really tidy.  As discussed I didn't look at the temp 2021 script

src/outputs/outputs_main.py
AnneONS: See comment below for context, but glad you noticed, maybe I should have left this in, I'll check.

src/site_apportionment/site_apportionment_main.py
AnneONS: Originally I was intending to do the imputation_backdata output at the end of estimation, so I wanted to do the clearing up we were going to do where we only did output of one data frame at the end of estimation. But I backtracked, and there are still some left over changes. In this case, I removed an incomplete comment.

JenCheshire: Great, so it's just the Args description that needs updating

PR #324, opened 2024-09-02
985 mor thresholds for each target var

AnneONS:
## Pull Request submission

In the config, users can specify an mor_threshold - if an imputation class is below this threshold, then Carry Forwards imputation will be used instead of MoR.

At present this threshold is set to 3, but it is recommended by Ana (methodology) to be 5.

At present, we only calculate this once for the 211 target variable but methodology suggest this ought to be done for each target variable in turn. This would mean also adding more fields for the imputation class size for each target variable, rather than just 211 as currently.

#### Closes or fixes

Closes #985

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/_version.py
```diff
@@ -1,2 +1 @@
-__version__ = "1.1.5"
-
+__version__ = "1.1.6"
```
zorge69, line 1: Agreed, let's test it in DAP as 1.1.6.

src/imputation/MoR.py
```diff
@@ -288,33 +289,29 @@ def group_calc_link(group, target_vars, config):
         link_vars ([string]): List of the linked variables.
         config (Dict): Confuration settings
     """
-    valid_group_size = True
+    threshold_num = get_threshold_value(config)
```
zorge69, line 318: In function group_calc_link, please add type hints and what it Returns

zorge69 (changes requested):
General impression: the code does what it is supposed to do, iterating through target vars.

Many functions in MoRT.py and sf_expansion need updating type hints and/or doc strings. Requested one change, relevant to this feature, group_calc_link - to update the docstring. Happy to approve once it's done.

src/imputation/MoR.py
AnneONS: done :-)

zorge69 (approved):
I can see the updated "Returns" :-)
The pipeline runs , run time 5 minutes 7 s.

PR #363, opened 2024-10-23
RDRP-1018

Ryan2Y79:
## Pull Request submission

Changes made to remove errors when the pipeline is run as a result of updating from `Python 3.6` to `Python 3.10`.

Unit test: Testing the behaviour of the modified scripts was limited by:
- No unit test available for `src/imputation/expansion_imputation.py` 
- Unit tests skipped for `src/imputation/tmi_imputation.py` & `src/expansion/sf_expansion.py`
- Two errors that were consistent between branches `develop` and `RDRP-1018` for src/imputation/test_MoR.py`
- Two failed tests that were consistent between branches `develop` and `RDRP-1018` for `src/staging/test_validation.py`

#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/estimation/apply_weights.py
```diff
@@ -33,9 +33,13 @@ def apply_weights(
     # TODO remove the forcing to numeric when validation fixed
```
AnneONS: we can remove this TODO now.

src/estimation/apply_weights.py
```diff
@@ -33,9 +33,13 @@ def apply_weights(
...
     # if the dataframe is for QA output, create new columns with the weights applied.
     if for_qa:
-        for col in cols_list:
-            df[col] = pd.to_numeric(df[col], errors="coerce")
-            df[f"{col}_estimated"] = round(df[col] * df["a_weight"], round_val)
+
+        estimated_cols = pd.concat(
+            [round(df[col] * df["a_weight"], round_val).rename(f"{col}_estimated") for col in cols_list],
```
AnneONS: we have pre-commit hooks that apply 88 lines, can you get these installed?

src/imputation/MoR.py
```diff
@@ -240,13 +240,13 @@ def calculate_growth_rates(current_df, prev_df, target_vars):
     prev_df = (
         prev_df[["reference", "imp_class", "imp_marker"] + target_vars]
         .groupby(["reference", "imp_class"])
-        .sum()
+        .sum(numeric_only=True)
     ).reset_index()
```
AnneONS, line 245: I'm wondering- if we used group_keys=False, would we need to have the reset_index?

src/imputation/imputation_helpers.py
```diff
@@ -262,10 +262,11 @@ def split_df_on_trim(df: pd.DataFrame, trim_bool_col: str) -> pd.DataFrame:
 
     if not df.empty:
         # TODO: remove this temporary fix to cast Nans to False
-        df[trim_bool_col].fillna(False, inplace=True)
+        df_copy = df.copy()
+        df_copy.loc[:, trim_bool_col].fillna(False, inplace=True)
```
AnneONS: I'm never sure when it's necessary to use .copy(), appart from to get rid of the warnings! 

For this pipeline, we prefer to avoid the inplace=True and explicitly set the updated column, though I can see it was used here.

What about this:
```suggestion
        df[trim_bool_col] = df.copy()[trim_bool_col].fillna(False)
   # or, if it still complains
        df.loc[:,trim_bool_col] = df.copy().loc[:,trim_bool_col].fillna(False)
```

src/imputation/sf_expansion.py
```diff
@@ -211,6 +211,11 @@ def run_sf_expansion(df: pd.DataFrame, config: dict) -> pd.DataFrame:
         threshold_num,
     )
 
+    # Set dtype of manual_trim column to bool before concatination
+    convert_dict = {'manual_trim': bool}
+    expanded_df = expanded_df.astype(convert_dict)
+    excluded_df = excluded_df.astype(convert_dict)
+
```
AnneONS, line 222: Interesting! was this as the result of a warning, or just something you noticed we'd need?

src/staging/validation.py
```diff
@@ -211,6 +211,10 @@ def combine_schemas_validate_full_df(
             survey_df[column] = survey_df[column].astype("float64", errors="ignore")
         elif dtypes[column] == "str":
             survey_df[column] = survey_df[column].astype("string")
+        elif pd.api.types.is_datetime64tz_dtype(survey_df[column]):
```
AnneONS, line 214: is this one of the types that by default our version of pandas uses for reading csvs? I'm wondering if we might need something more general for if, eg, we have different versions for python/ pandas. But if it works, it works!

src/imputation/tmi_imputation.py
```diff
@@ -563,6 +563,12 @@ def run_tmi(
     imputed_only_df = imputed_only_df.sort_values(
         ["formtype", "imp_class"], ascending=True
     ).reset_index(drop=True)
+    
+    # Set dtype of manual_trim column to bool before concatination
+    convert_dict = {'manual_trim': bool}
+    full_qa_df = full_qa_df.astype(convert_dict)
+    imputed_only_df = imputed_only_df.astype(convert_dict)
```
AnneONS, line 574: I've done some analyis with copilot and I think all these cols are bools and are in those 2 dataframes:

```suggestion
    bool_like_cols =  ['manual_trim', 'empty_pgsic_group', 'empty_pg_group', '211_trim', '305_trim']
```

AnneONS: do you think manual_trim is a special case?

AnneONS (commented):
HI Ryan, 
Glad you were able to make all these changes fairly quicly! I have some discussion points for you here.

src/staging/validation.py
AnneONS: can you put in a comment ?

src/estimation/apply_weights.py
Ryan2Y79: DONE

src/estimation/apply_weights.py
Ryan2Y79: DONE: I followed the R&D user guidance.

src/imputation/MoR.py
Ryan2Y79: According to Co-pilot we can. It would look like this:

```Python
current_df = (
        current_df[["reference", "imp_class", "imp_marker"] + target_vars]
        .groupby(["reference", "imp_class"], group_keys=False)
        .sum(numeric_only=True)
    )
 ```
 
 Do you want this new `group_keys=False` version?

File: src/imputation/imputation_helpers.py (hunk 4 above)
Comment (Ryan2Y79) [reply src/imputation/imputation_helpers.py]: Noted on the avoidance of `inplace=True` for this pipeline. The implementation of `inplace=True` for this line was implemented a previous developer and was approved November last year, therefore I assumed its used was tolerated. 

The first suggestion causes the pipeline to crash and the second suggestion also kills the pipeline. I have modified the line from this:

```Python
df_copy.loc[:, trim_bool_col].fillna(False, inplace=True)
```
to:
``` Python
df_copy.loc[:, trim_bool_col] = df_copy.loc[:, trim_bool_col].fillna(False)
```

The pipeline works with this latest version and satisfies the removal of `inplace=True`.

src/imputation/sf_expansion.py
Ryan2Y79: This was as a results of a warning. At this stage of a pipeline run, `manual_trim` dtype is `object`, however the pipeline expects the dtype to be `bool`. Based on your comment for `src/imputation/tmi_imputation.py`, I have updated `convert_dict` to include additional columns for conversion to bools (belt and braces approach taken).

src/imputation/tmi_imputation.py
Ryan2Y79: Yes, `manual_trim` dtype at this stage in the pipeline is as an `object`, whereas it should be set as a `bool`. I have added the other suggested columns to the `convert_dict` and the pipeline runs. NOTE: I also updated `src/imputation/sf_expansion.py` line 215 `convert_dict` to include  those columns (belt and braces approach taken).

src/staging/validation.py
Ryan2Y79: I updated the comment to reflect further what this has been added.

src/imputation/MoR.py
AnneONS: let's leave it for now- we can't risk any changes that would change outputs this close to publication. But when we create our unit tests for this code, which is something we'll hopefully do before long, we can investigate properly using the unit tests.

AnneONS: I say this because I think it's the index that changes, and Dylan found this caused an unexpected bug further down the pipeline. So I want to test it thoroughly rather than just seeing if the numbers initially look the same.

src/imputation/imputation_helpers.py
AnneONS: thanks for checking all those things.

src/imputation/tmi_imputation.py
AnneONS: yes, I think that's best.

src/staging/validation.py
AnneONS: that's helpful, thanks.

AnneONS (approved):

PR #368, opened 2024-11-05
RDRP-1036 update make dirs for network and s3

Ryan2Y79:
## Pull Request submission

The current `make_dirs_main.py` is used to make directories within S3 buckets for CDP, and not for local network drives. I have created a duplicate version of `make_dirs_main.py`, and refactored the script to create directories within the R: drive, and to meet the requirements of the task ticket.

#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


helpers/make_network_dirs_main.py
```diff
@@ -0,0 +1,171 @@
...
+    tree (dict) converted to list of lists by function src.utils.helpers.tree_to_list.
+    make_dir function is called for item in the list of directories.
+
+    """
+
+    root = "BERD Results System Development 2023/DAP_emulation"
+
+    tree = {
+        "2022_surveys": {
+            "PNP": {
```
AnneONS: Is there a way to do this where you don't have to hard code survey (PNP / BERD) or year, and you can put these in at the top?

Ryan2Y79: I gave addressed this comment by adding a loop into the function that iterates through years and survey types.

helpers/make_network_dirs_main.py
```diff
@@ -0,0 +1,118 @@
...
+    root = "BERD Results System Development 2023/DAP_emulation"
+
+    years = ["2022", "2023"]
+    surveys = ["BERD", "PNP",]
+
+    for year in years:
+        for survey in surveys:
+
+            tree = {
+                f"{year}_surveys": {
```
AnneONS, line 41: line 45 can be outside the loop so it isn't done each time.

helpers/make_network_dirs_main.py
```diff
@@ -0,0 +1,118 @@
...
+                            "output_intram_uk_itl2": {},
+                            "output_long_form": {},
+                            "output_ni_sas": {},
+                            "output_short_form": {},
+                            "output_status_filtered_qa": {},
+                            "output_tau": {},
+                        },
+                    },
+                    "mappers": {
+                        "v1": {},
```
AnneONS, line 103: the mappers dictionary can also be outside the survey_type loop, it is just created once per year.

helpers/make_network_dirs_main.py
```diff
@@ -0,0 +1,118 @@
+"""Script that creates all directories"""
+import os
+import sys
+sys.path.append("D:/research-and-development")
```
AnneONS: instead of this line (it will be different for all of us), move line 9 above here and use this instead in the append.
`my_wd = os.getcsd()`

helpers/make_network_dirs_main.py
```diff
@@ -0,0 +1,118 @@
...
+from src.utils.helpers import tree_to_list
+from src.utils.local_file_mods import rd_file_exists
+
+# Change to the project repository location
+my_wd = os.getcwd()
+my_repo = "R:\\"
```
AnneONS: Would it make more sense to move "root" (line 37) here, and then call the "tree_to_list" function with the default empty string for "prefix"?

Ryan2Y79: I have addressed this comment. Thank you.

helpers/make_network_dirs_main.py
Ryan2Y79: this suggestion makes total sense. I have addressed this in the script. Thanks.

helpers/make_network_dirs_main.py
Ryan2Y79: Implementing this suggestion would actually add complexity to the script. One would have to 1. check if the folder are already created 2. create an `if` `else` statement to create or not create a new folder dependant on if the folder does/doesn't exist. Considering this script is not part of the pipeline, and will only be used once in a blue moon, the 
current genuine inefficiency has zero impact. I suggest we keep the current iteration.

AnneONS (approved):

PR #376, opened 2024-11-13
RDRP- 997 Replace toml.load with tomli.load

woodac:
Replaced toml.load with tomli.load, as it is faster and is supported by Python 3.11.

Added toml decode error exception as a co-pilot suggestion, not 100% its effective as it is not returning anything at the end of the statement.

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/outputs/export_files.py
```diff
@@ -6,6 +6,7 @@
 import logging
 from datetime import datetime
 import toml
+import tomli
```
adriano-lopresti, line 8: the toml package is no longer used after switching to tomli. "import toml" can therefore be removed

src/staging/validation.py
```diff
@@ -1,5 +1,6 @@
 import os
 import toml
+import tomli
```
adriano-lopresti, line 2: the toml package is no longer used after switching to tomli. "import toml" can therefore be removed

src/utils/helpers.py
```diff
@@ -1,6 +1,7 @@
 """Define helper functions to be used throughout the pipeline.."""
 import yaml
 import toml
+import tomli
```
adriano-lopresti, line 3: the toml package is no longer used after switching to tomli. "import toml" can therefore be removed

src/outputs/export_files.py
```diff
@@ -49,7 +50,7 @@ def get_schema_headers(config: dict):
 
     # Get the headers for each
     schema_headers_dict = {
-        output_name: toml.load(path) for output_name, path in schema_paths.items()
+        output_name: tomli.load(path) for output_name, path in schema_paths.items()
```
adriano-lopresti: unlike in "toml" , in "tomli" method load can't take in a string: it takes a binary (the one you would get with open(..) as f.
In order to read a string try "loads" (with an s) instead.

src/staging/validation.py
```diff
@@ -26,17 +27,20 @@ def load_schema(file_path: str = "./config/contributors_schema.toml") -> dict:
 
     # Check if Data_Schema.toml exists
     if file_exists:
-        # Load toml data schema into dictionary if toml file exists
-        toml_dict = toml.load(file_path)
+        try:
+            # Open the file and load toml data schema into dictionary
+            with open(file_path, 'rb') as file:
+                toml_dict = tomli.load(file)
```
adriano-lopresti, line 32: this should be fine!

src/utils/helpers.py
```diff
@@ -44,7 +45,7 @@ def user_config_reader(configfile: str = user_config_path) -> dict:
         {'title': 'TOML Example config', 'period': {'start_period':
         datetime.date(1990, 10, 10), 'end_period': datetime.date(2000, 10, 5)}}
     """
-    toml_dict = toml.load(configfile)
+    toml_dict = tomli.load(configfile)
```
adriano-lopresti: unlike in "toml" , in "tomli" method load can't take in a string: it takes a binary (the one you would get with open(..) as f.
In order to read a string try "loads" (with an s) instead.

src/staging/validation.py
```diff
@@ -26,17 +26,20 @@ def load_schema(file_path: str = "./config/contributors_schema.toml") -> dict:
...
-        # Return False if file does not exist
+        # Return None if file does not exist
         ValidationLogger.warning(
             "Validation schema does not exist! Path may be incorrect"
         )
-        return file_exists
-
-    return toml_dict
-
+        return None
```
Ryan2Y79, line 42: Forgive my ignorance, but the function previously returned `toml_dict`, but in these changes the function returns `None` in both scenarios. Is this correct?

Ryan2Y79 (commented):
Just one question to address. Cheers.

src/staging/validation.py
AnneONS: @Ryan2Y79 no, look at line 33, we return toml_dict there

AnneONS (approved):

PR #387, opened 2024-12-03
RDRP-1045_pnp_add_osmotherly_key_area_to_current

Ryan2Y79:
## Pull Request submission

#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/imputation/imputation_helpers.py
```diff
@@ -440,3 +484,75 @@ def create_new_backdata(backdata: pd.DataFrame, config) -> pd.DataFrame:
...
+def identify_key_business(df):
+    """ Function to identify the key business using reference column & key
+    busineses lookup table.
+    Args:
+        df (pd.DataFrame): The dataframe to identify the key business columns.
+    Return:
+        df (pd.DataFrame): The dataframe with the identified key business columns.
+    """
+    # get key businesses
+    rootpath = "R:/BERD Results System Development 2023/DAP_emulation/2021_surveys/PNP/"
```
AnneONS: we don't hard code paths, it's not best practice and it won't work in the prod environment :-)

src/imputation/imputation_main.py
```diff
@@ -50,6 +50,15 @@ def run_imputation(
     Returns:
         pd.DataFrame: dataframe with the imputed columns updated
     """
+    # add osmotherly & key businesses columns to PNP data
+    if config["survey"]["survey_type"] == "PNP":
+        df = hlp.identify_key_business(df)
+        df = hlp.identify_osmotherly_businesses(df)
+        df = hlp.add_area_column(df)
```
AnneONS: I suggest this all moves to the mapping module.

src/utils/helpers.py
```diff
@@ -250,3 +250,30 @@ def filename_amender(filename, config):
...
+        schema_dict.update({'pnp_key': {'old_name': 'pnp_key',
+                                        'Deduced_Data Type': 'string'}})
+        schema_dict.update({'osmotherly': {'old_name': 'osmotherly',
+                                           'Deduced_Data Type': 'string'}})
+        schema_dict.update({'area': {'old_name': 'area',
+                                     'Deduced_Data Type': 'string'}})
+    else:
+        None
+
+    return schema_dict
```
AnneONS: This might be useful, we will see.

src/imputation/imputation_main.py
```diff
@@ -141,6 +150,7 @@ def run_imputation(
         # create trimming qa dataframe with required columns from schema
         schema_path = config["schema_paths"]["manual_trimming_schema"]
         schema_dict = load_schema(schema_path)
+        schema_dict = add_additional_columns_to_schema(schema_dict, config)
         trimming_qa_output = create_output_df(qa_df, schema_dict)
```
AnneONS: are these extra columns actually needed for the trimming_qa_output?

src/imputation/imputation_main.py
```diff
@@ -50,6 +50,15 @@ def run_imputation(
...
     """
+    # add osmotherly & key businesses columns to PNP data
+    if config["survey"]["survey_type"] == "PNP":
+        df = hlp.identify_key_business(df)
+        df = hlp.identify_osmotherly_businesses(df)
+        df = hlp.add_area_column(df)
+        df = hlp.create_imp_class_col(df,
+                                  ["pnp_key", "area"],
+                                  use_osmotherly=True,
+                                  use_cellno=False)
```
AnneONS: If we create the imputation class here, we'll get stuck when the MoR code tries to do it again- so maybe we could move this code to where the create_imp_class_col is created in the MoR script.

AnneONS (commented):
There is still a lot to do before this code will be ready to run.

We should discuss.

src/imputation/imputation_main.py
Ryan2Y79: Done :)

src/imputation/imputation_main.py
Ryan2Y79: You are correct in that `trimming_qa_output` will work without these extra columns. However without these extra columns being added to the `scheme_dict` the columns will not be added to the output and thus the QA files will be incomplete when PNP mode is run.

src/imputation/imputation_main.py
Ryan2Y79: Your observation here is correct only if both parameters within `create_imp_class_col`, `use_osmotherly` & `use_cellno` are both set to `True` at the same time. To circumvent this, I have set `use_osmotherly` to default as `False`. 

However you have highlighted something to me, and that is why should be be creating the `imp_class` in two separate scripts? It would be better to create the `imp_class` in one place, and one place only. Furthermore we should introduce a logic check so that  `use_osmotherly` & `use_cellno`  cannot be `True` at the same time.

Ryan2Y79 (commented):
mappers moved to mapping module, and code is now platform agnostic.

src/imputation/imputation_main.py
AnneONS: but the schema_dict is only for the manual_trimming_schema, so it's not needed.

AnneONS: I suggest the only time we might need to see those extra 3 cols is in the mapping QA output, we don't really need them after we create the imputation class, as you noted. What do you think?

src/mapping/mapping_main.py
```diff
@@ -96,6 +96,14 @@ def run_mapping(
     else:
         MappingMainLogger.info(f"Reference list not updated for survey year {year}.")
 
+    # Add custom mappers if PNP is true for later imp_class column creation
+    if config["survey"]["survey_type"] == "PNP":
+        full_responses = hlp.identify_osmotherly_key_area(
```
AnneONS: pretty good, and I like having all of this together in one place.
Just a couple of suggestions, can the new scripts you added to mapping_helpers be put in another separate script called pnp_mapping.py or something? Just like we did for cellno, itl, pg, ultfoc. Don't spend too long on it though, as we're not 100% sure exactly how we'll be doing this.

AnneONS (commented):
I like the way you code :-) Just a few things, then let's move on.

src/mapping/mapping_main.py
Ryan2Y79: Done :)

src/mapping/mapping_main.py
```diff
@@ -8,6 +8,7 @@
 from src.mapping.ultfoc_mapping import join_fgn_ownership
 from src.mapping.cellno_mapping import validate_join_cellno_mapper
 from src.mapping.itl_mapping import join_itl_regions
+from src.mapping.pnp_mapping import identify_osmotherly_key_area
```
AnneONS, line 12: this file doesn't exist, so we can't run the pipeline.

AnneONS (approved):
More changes were needed to make the pipeline run, but I have now fixed these.

PR #388, opened 2024-12-03
RDRP-1065_pnp_config_logic

woodac:
Create a new function as part of config validation that stops the pipeline, or creates a warning, if config settings are not compatible for a PNP run.

For example:

If load_ni_data is true, print a message and stop the pipeline (and remove the bit of code that currently does this in the NI module)
If the output settings for output_ni_sas, or the itl_uk or pg_uk, log a warning and update these settings to False but don't stop the pipeline
If any other settings for outliers or ni are set to true, warn and update to False
If any qa outputs for outliers or estimation are true, warn and update to False.
The suggested place for this new pnp_config_check function is in src.utils.config.py, and it could be called by validate_config( ) - check that this will work ok.
#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/pipeline.py
```diff
@@ -45,6 +45,12 @@ def run_pipeline(user_config_path, dev_config_path):
     # validate the filenames in the config
     config = filename_validation(config)
 
+    # Check validate survey config
+    config = validate_survey_config(config)
+
+    # Check if PNP is set and config is valid
+    config = validate_pnp_config(config)
+
```
AnneONS, line 46: to keep pipeline.py as simple as possible, could these two new functions be combined with filename_validation, do you think? You might like to give filename_validation a new name. 

By the way, the reason we can't do this validation when we actually set up the config is because we need to do it after the logger has been initialised.

src/utils/config.py
```diff
@@ -402,3 +402,68 @@ def validate_construction_config_settings(config):
...
+ 
+        values = [config.get("global", {}).get(setting, True)
+                  for setting in incompatible_settings]
+ 
+        if any(values):
+             print(
+                "WARNING: PNP is set. The following settings are not compatible with PNP: "
+                "output_ni_full_responses, output_mapping_ni_qa, output_short_form, load_ni_data,"
+                "run_ni_construction, output_ni_sas, output_intram_by_pg_gb and output_outlier_qa."
+                " User config will be updated to False."
```
AnneONS: could you have this so it only outputs the items they actually chose? Do you think that would be useful?

AnneONS (commented):
just a brief look so far, it all looks good but I have a couple of suggestions

src/utils/config.py
woodac: Updated code so that it prints the outputs that it has changed :)

src/utils/config.py
```diff
@@ -402,3 +402,66 @@ def validate_construction_config_settings(config):
                 "If running NI construction, a NI construction file path must be"
                 " provided."
             )
+
+def validate_survey_config(config: dict) -> dict:
+    """Validate the configuration settings.
```
AnneONS: can this be more specific, say we're doing PNP validation

src/utils/config.py
```diff
@@ -402,3 +402,66 @@ def validate_construction_config_settings(config):
...
+    # Check if PNP is set
+    if config["survey"]["survey_type"] == "PNP":
+        # List of values not compatible with PNP
+        incompatible_settings = ["output_ni_full_responses",
+                                 "output_mapping_ni_qa",
+                                 "output_short_form",
+                                 "run_ni_construction",
+                                 "output_ni_sas",
+                                 "output_intram_by_pg_gb",
+                                 "output_outlier_qa"]
```
AnneONS: Has this been run through the pre-commits? I'd expect it to update the indenting.

src/utils/config.py
```diff
@@ -402,3 +402,66 @@ def validate_construction_config_settings(config):
...
+        
+        result = []
+        for setting in incompatible_settings:
+            # Check if setting is True
+            if config.get("global", {}).get(setting, True):
+                # Add setting to result list
+                result.append(setting)
+                # Print warning message with setting that has bee updated.
+                print( "WARNING: PNP is set. The following settings are not compatible with PNP: "
+                 f"{result}" "User config will be updated to False.")
```
AnneONS: this needs to go outside the for loop

src/utils/config.py
```diff
@@ -402,3 +402,66 @@ def validate_construction_config_settings(config):
...
+                #Update to False
+                for setting in incompatible_settings:
+                  config["global"][setting] = False
```
AnneONS: you don't need an extra for loop in here, it can just be done as part of the orig for loop, when the if is True.

AnneONS (commented):
A good start, I like how you've constructed your list of settings.

src/utils/config.py
woodac: You are right, pre-commits weren't working. Got them working so now resolved.

src/utils/config.py
woodac: Agreed, refactored to remove nested loop.

src/utils/config.py
woodac: Done :)

src/utils/config.py
```diff
@@ -403,3 +403,84 @@ def validate_construction_config_settings(config):
...
+    # Check if PNP is set
+    if config["survey"]["survey_type"] == "PNP":
+        # List of values not compatible with PNP
+        incompatible_settings = [
+            "output_ni_full_responses",
+            "output_mapping_ni_qa",
+            "output_short_form",
+            "run_ni_construction",
+            "output_ni_sas",
+            "output_intram_by_pg_gb",
```
AnneONS: this should be intram_by_pg_uk, not gb, and we also need intram_uk_itl

src/utils/config.py
```diff
@@ -403,3 +403,84 @@ def validate_construction_config_settings(config):
...
+        if any(values):
+            raise ValueError(
+                "Error: PNP is set. Northern Ireland data cannot be run. Please update "
+                "the user config."
+            )
+    return config
+
+
+def validate_pnp_config(config: dict) -> None:
+    """
```
AnneONS, line 436: please update the docsrting

AnneONS (commented):
Just the docstrings, and you need to look for the "uk" outputs which include NI so can't be done with PNP

src/utils/config.py
```diff
@@ -403,3 +403,91 @@ def validate_construction_config_settings(config):
...
+        incompatible_settings = [
+            "output_ni_full_responses",
+            "output_mapping_ni_qa",
+            "output_short_form",
+            "run_ni_construction",
+            "output_ni_sas",
+            "output_intram_by_pg_uk",
+            "output_outlier_qa",
+            "output_intram_uk_itl1",
+            "output_intram_uk_itl2",
```
AnneONS: although we have outputs with the names _itl1 and itl2, the actual setting covers both together, and its just output_intram_uk_itl

woodac: Yes, sorry Anne, I was looking at the output files yesterday to put them in as that, instead of what is in the config. I've updated, so should now be correct.

AnneONS (approved):
ready to go!

PR #397, opened 2025-01-09
RDRP-1090_refactor_tmi_functions

woodac:
Removing unnecessary functions from theTMI_imputation.py script : apply_to_original and filter_by_column_content. Where these appear at any point in the codebase, remove them and just apply the python inline as this is clearer.

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/imputation/impute_civ_def.py
```diff
@@ -248,10 +251,18 @@ def apply_civdev_imputation(
...
-        tmi.apply_to_original(class_group_df, filtered_df2)
-    tmi.apply_to_original(filtered_df2, to_impute_df)
+        # Apply changes from class_group_df to filtered_df2
+        filtered_df2.update(class_group_df)
+
+    # Apply changes from filtered_df2 to to_impute_df
+    to_impute_df.update(filtered_df2)
+
+    # Apply changes from to_impute_df to df
+    df.update(to_impute_df)
```
AnneONS, line 261: there are a couple of places where you are doing a .update and then on the next line assigning to a new variable, to keep the variable names as they were before.
Instead, could we have the update and the rename on the same line? eg
updated_df = df.update(to_impute_df)

src/imputation/tmi_imputation.py
```diff
@@ -356,9 +348,10 @@ def apply_tmi(
                 imp_class_df["imp_marker"] = "No mean found"
 
             # Apply changes to copy_df
-            final_df = apply_to_original(imp_class_df, filtered_df)
-
-    final_df = apply_to_original(final_df, df)
+            filtered_df.update(imp_class_df)
+        final_df = filtered_df
+        df.update(final_df)
+    final_df = df
```
AnneONS: again, could we do the rename and update in the same line? Would this work the same?

AnneONS (commented):
looks good so far! Please can you check that the imputation qa output is the same in your changes as in the develop branch? In helpers we have a regression_imputation script that is designed to help compare output of 2 imputation_qa files

src/imputation/tmi_imputation.py
```diff
@@ -356,11 +340,11 @@ def apply_tmi(
                 imp_class_df["imp_marker"] = "No mean found"
 
             # Apply changes to copy_df
-            final_df = apply_to_original(imp_class_df, filtered_df)
+            filtered_df.update(imp_class_df)
 
-    final_df = apply_to_original(final_df, df)
+        df.update(filtered_df)
```
AnneONS: this change is not quite the same  - the new line 345 is indented one tab more than the original 361.
Although it might not make a difference in practice , I think we'd better change it (though we'd need to make sure the output was as expected)

AnneONS (commented):
Could you address this one comment for me? Otherwise it looks great!

src/imputation/tmi_imputation.py
woodac: Removed the extra indent in new line 345 to be line with original 361 and have re-run the test which has produced the output we were expecting.

AnneONS (approved):

PR #406, opened 2025-01-22
RDRP 1141- adding_uni_employment_column

woodac:
Adding the “uni_employment” column, that needs to be mapped to each cell in the survey data at the same time that the “Uni_Count” column is mapped from the cellno mapper in the mapping module (AC: an extra column in the mapping_qa output)

Step needed:

src\mapping\cellno_mapping.py line 30 changed to include "uni_employment" eg.

cellno_df = cellno_df.copy()[["cell_no", "UNI_Count", "uni_employment"]]

Validation check that the mapping is correct for "uni_employment" with the same method as "uni_count"

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/mapping/mapping_helpers.py
```diff
@@ -6,26 +6,22 @@
...
 
     """
-    if mapper_df[col1].isnull().any():
-        raise ValueError(f"{mapper_name} mapper contains null values in {col1}.")
-    if mapper_df[col2].isnull().any():
-        raise ValueError(f"{mapper_name} mapper contains null values in {col2}.")
+    for col in mapper_df.columns:
+        if mapper_df[col].isnull().any():
+            raise ValueError(f"{mapper_name} mapper contains null values in {col}.")
+    return mapper_df
```
AnneONS: I can see how it would make sense to check all columns for nulls in a mapper.
This would work for our pipeline, as we are only keeping the columns we use.

However, do you think it might be more useful if we let the user choose which  cols to validate, to be more general? Maybe pass a list as an option, and default to df.columns otherwise?

src/staging/staging_helpers.py
```diff
@@ -132,6 +133,9 @@ def load_validate_mapper(
     # Validate the DataFrame against the schema
     val.validate_data_with_schema(mapper_df, schema_path)
 
+    # Perform null checks on the mapper DataFrame
+    mapper_df = mapper_null_checks(mapper_df, mapper_name)
```
AnneONS: good addition to our validation

AnneONS (commented):
looks good! Can we also have an (updated) unit test for the validation function you have changed?

src/mapping/mapping_helpers.py
```diff
@@ -6,26 +6,63 @@
...
 
     """
-    if mapper_df[col1].isnull().any():
-        raise ValueError(f"{mapper_name} mapper contains null values in {col1}.")
-    if mapper_df[col2].isnull().any():
-        raise ValueError(f"{mapper_name} mapper contains null values in {col2}.")
+    # Create list to add columns to validate
+    validate_cols = ["uni_count", "uni_employment"]
+
+    # Mappers to ignore (expected NULL values)
```
woodac: itl mapper- column E- ITL321NM, has 3 null values that were stopping the pipeline. May need to review this mapper, if this is expected or not.

src/mapping/mapping_helpers.py
```diff
@@ -6,26 +6,63 @@
...
+    for col in columns:
+        if col not in df.columns:
+            raise Warning(f"Column {col} not found in DataFrame.")
+
+
 def mapper_null_checks(
-    mapper_df: pd.DataFrame, mapper_name: str, col1: str, col2: str
-) -> None:
+    mapper_df: pd.DataFrame, mapper_name: str, validate_cols=False
+) -> pd.DataFrame:
```
AnneONS: what would happen if we did want to validate a few cols, not all of them? We'd need to be passing a list. So the final arg would need to be a list. At the moment, it looks like it could also be a bool. 
Python I think allows you to have an argument that swaps type, but it's dodgy behaviour, and if we use pydantic, this wouldn't work.
So the default behaviour could be that the list is empty, ie [ ]  - then you could have if [ ] is passed you do all (which is not that neat actually... can you think of a neater way? Maybe you could have
validate_cols: list=None
then it would know it was a list and you could check "if None then do all"

src/mapping/mapping_helpers.py
```diff
@@ -6,26 +6,63 @@
...
     """
-    Perform null checks on two columns of a mapper DataFrame.
+    Perform null checks on all columns of a mapper DataFrame.
```
AnneONS: we might not be doing all, we want "selected" (defaulting to all)

src/mapping/mapping_helpers.py
AnneONS: aha!! In that case, we don't want to validate all columns for the itl mapper, but we do for the cellno.

AnneONS: in this function, you are hardcoding the names of columns, we don't want to do that- we want to pass in a list of cols to validate, or do all of them if no list passed.

AnneONS (commented):
A suggested change- explicitly pass a list of columns to check so the function can be general for any type of mapper and hard-coded names not needed.

src/mapping/mapping_helpers.py
```diff
@@ -7,25 +7,27 @@
 
 
 def mapper_null_checks(
-    mapper_df: pd.DataFrame, mapper_name: str, col1: str, col2: str
+    mapper_df: pd.DataFrame, mapper_name: str, validate_cols: list = None
 ) -> None:
     """
-    Perform null checks on two columns of a mapper DataFrame.
+    Perform null checks on selected columns of a mapper DataFrame.
```
AnneONS: this is a really small point, but in general we have the """ at the top on the same line as the writing.

src/mapping/mapping_helpers.py
```diff
@@ -7,25 +7,27 @@
...
-        raise ValueError(f"{mapper_name} mapper contains null values in {col1}.")
-    if mapper_df[col2].isnull().any():
-        raise ValueError(f"{mapper_name} mapper contains null values in {col2}.")
+    if validate_cols is None:
+        validate_cols = mapper_df.columns.tolist()
+    # Check for NULL values of all columns in the list
+    for col in validate_cols:
+        if mapper_df[col].isnull().any():
+            # Raise an error
+            raise ValueError(f"{mapper_name} contains unexpected null values in {col}.")
```
AnneONS: This is really great! Just one suggestion, I think the pipeline will end as soon as it finds a single column with nulls. Elsewhere we've got examples where we first get a dataframe or a list of the cols with an issue, and then if that df / list isn't empty, we output a message that includes all the relevant info. (eg, see lines 64-65 below)

src/mapping/mapping_main.py
```diff
@@ -53,6 +53,16 @@ def run_mapping(
...
+            "LAD20CD",
+            "LAD20NM",
+            "LAU121CD",
+            "LAU121NM",
+            "ITL221CD",
+            "ITL221NM",
+            "ITL121CD",
+            "ITL121NM",
+        ],
     )
```
AnneONS, line 60: We don't actually use all these columns, if you look at lines 40-41 in the itl mapper you can see we just use some of them:
    # next join the itl mapper to add the region columns
    gb_itl_col = config["mappers"]["gb_itl"]
    geo_cols = [gb_itl_col] + config["mappers"]["geo_cols"]

maybe create a list to collect those cols in a variable and use that. It also means we're doing less hardcoding- if these names changed, we could get hte latest version from the config.

AnneONS (commented):
nearly there! Great work

AnneONS (commented):
thanks

src/mapping/mapping_helpers.py
```diff
@@ -7,25 +7,39 @@
...
-    if mapper_df[col1].isnull().any():
-        raise ValueError(f"{mapper_name} mapper contains null values in {col1}.")
-    if mapper_df[col2].isnull().any():
-        raise ValueError(f"{mapper_name} mapper contains null values in {col2}.")
+    if validate_cols is None:
+        validate_cols = mapper_df.columns.tolist()
+    # Check for NULL values of all columns in the list
+    for col in validate_cols:
+        if mapper_df[col].isnull().any():
+            if not mapper_df.empty:
```
AnneONS: I'm not sure if we need this line, do we?

src/mapping/mapping_helpers.py
```diff
@@ -7,25 +7,39 @@
...
+                msg = (
+                    f"{mapper_name} contains unexpected null values in {col}."
+                    f"The following {col} values are not in the {mapper_name} mapper: "
+                    f"{mapper_df[col].unique()}"
+                )
+                if warn:
+                    MappingLogger.warning(msg)
+                else:
+                    raise ValueError(msg)
```
AnneONS: This would still stop after the first one if we have "raise". Can you think of a way to run through all the columns and then put the message covering all errors at the end?

src/mapping/mapping_helpers.py
AnneONS: I suppose it might be good to check if the df is empty (really we should have done this when reading it in...). But we don't need to check it each time we go through the list, maybe do it at the beginning.

AnneONS (commented):
it's nearly ready :-)

adriano-lopresti (approved):
Happy for this request to be merged

PR #415, opened 2025-01-28
RDRP- 1143_function_to_sum_emp_sample

woodac:
This function calculates the sum of column "employment" for the filtered dataframe to create lower_e.

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)

### DIFF
--- a/src/estimation/calculate_weights.py
+++ b/src/estimation/calculate_weights.py
@@ -54,8 +54,9 @@ def calc_lower_e(df: pd.DataFrame, emp_col: str = "711") -> dict:
     # Check if any of the key cols are missing
     cols = set(df.columns)
     required_cols = ["reference", "cellnumber", emp_col]
-    if not all(col in cols for col in required_cols):
-        raise ValueError(" One or more of the required columns are missing.")
+    for col in required_cols:
+        if col not in cols:
+            raise ValueError(f"{col} missing.")
 
     # Sum employment for each cellnumber
     e = df.groupby("cellnumber")[emp_col].sum()
--- a/src/estimation/calculate_weights.py
+++ b/src/estimation/calculate_weights.py
@@ -16,46 +16,32 @@ def create_estimation_filter(df: pd.DataFrame) -> pd.Series:
     return estimation_filter
 
 
-def calc_lower_n(df: pd.DataFrame, exp_col: str = "709") -> dict:
+def calc_lower_n(df: pd.DataFrame) -> int:
     """Calculates 'n' which is a number of
     unique RU references in the filtered dataset.
 
     Args:
         df (pd.DataFrame): The input dataframe which contains survey data,
             including expenditure data
-        exp_col (str): An appropriate column to count n
-
     Returns:
         int: The number of unique references.
     """
-
-    # Check if any of the key cols are missing
-    cols = set(df.columns)
-    if not ("reference" in cols) & (exp_col in cols):
-        raise ValueError(f"'reference' or {exp_col} missing.")
-
     # Count the records
     n = df["reference"].nunique()
 
     return n
 
 
-def calc_lower_e(df: pd.DataFrame, emp_col: str = "711") -> int:
+def calc_lower_e(df: pd.DataFrame) -> int:
     """Calculates 'e' which is a sum of
     IDBR employment data in the filtered dataset.
 
     Args:
         df (pd.DatatFrame): The input dataframe which contains survey data,
             including IDBR employment data.
-        emp_col (str): An appropriate column to store sum employment data.
     Returns:
-        int: The sum of IDBR employment of sampled data within each cell.
+        int: The sum of IDBR employment of sampled.
     """
-    # Check if any of the key cols are missing
-    cols = set(df.columns)
-    if not (("employment" in cols) & (emp_col in cols)):
-        raise ValueError(f"employment or {emp_col} missing.")
-
     # Sum employment for each cellnumber
     e = df["employment"].sum()
 
--- a/src/estimation/calculate_weights.py
+++ b/src/estimation/calculate_weights.py
@@ -54,7 +54,7 @@ def calc_lower_e(df: pd.DataFrame, emp_col: str = "711") -> dict:
     # Check if any of the key cols are missing
     cols = set(df.columns)
     if not ("employment" in cols) & (emp_col in cols):
-        raise ValueError(f"'reference' or {emp_col} missing.")
+        raise ValueError(f"employment' or {emp_col} missing.")
 
     # Sum employment for each cellnumber
     e = df["employment"].sum()
--- a/src/estimation/calculate_weights.py
+++ b/src/estimation/calculate_weights.py
@@ -40,6 +40,35 @@ def calc_lower_n(df: pd.DataFrame, exp_col: str = "709") -> dict:
     return n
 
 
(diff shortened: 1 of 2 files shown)


src/estimation/calculate_weights.py
```diff
@@ -40,6 +40,35 @@ def calc_lower_n(df: pd.DataFrame, exp_col: str = "709") -> dict:
...
+
+    # Check if any of the key cols are missing
+    # cols = set(df.columns)
+    # TO DO: Check columns are present and valid responses
+    # 711 = Total Full Time Equivalent
+
+    # Get unique references
+    unique_refs = df["reference"].unique()
+
+    # Group by cellnumber
```
AnneONS: I don't think it would work to groupby here, as df["reference"].unique() is a list

src/estimation/calculate_weights.py
```diff
@@ -40,6 +40,35 @@ def calc_lower_n(df: pd.DataFrame, exp_col: str = "709") -> dict:
...
+    unique_refs = unique_refs.groupby("cellnumber")
+
+    # Sum employment for each cell
+    e = unique_refs["cellnumber"]["employment"].sum()
```
AnneONS: you can assume that we've already done the groupby. What you need to do is assume you've been passed the filtered dataframe for just one group, and simply return a number which is the sum.

AnneONS (commented):
good start, I've put in a couple of pointers.

src/estimation/calculate_weights.py
```diff
@@ -40,6 +40,28 @@ def calc_lower_n(df: pd.DataFrame, exp_col: str = "709") -> dict:
...
+    Args:
+        df (pd.DatatFrame): The input dataframe which contains survey data,
+            including IDBR employment data.
+        emp_col (str): An appropriate column to store sum employment data.
+    Returns:
+        int: The sum of IDBR employment of sampled data within each cell.
+    """
+    # Check if any of the key cols are missing
+    cols = set(df.columns)
+    if not ("employment" in cols) & (emp_col in cols):
```
adriano-lopresti: just out of clarity, can I ask you to add some parentheses arond this logic? is it to be read  "(not 1) and 2" "not ( 1 & 2 )", i.e.
(not ("employment" in cols) ) & (emp_col in cols)
or 
not (("employment" in cols)  & (emp_col in cols))
I think it's the first, but better to be unnecessarily explicit to avoid confusion

src/estimation/calculate_weights.py
```diff
@@ -40,6 +40,28 @@ def calc_lower_n(df: pd.DataFrame, exp_col: str = "709") -> dict:
     return n
 
 
+def calc_lower_e(df: pd.DataFrame, emp_col: str = "711") -> dict:
```
adriano-lopresti: why this definition seems to say that the function returns a dictionary whilst it actually returns an integer?

tests/test_estimation/test_calculate_weights.py
```diff
@@ -156,6 +156,70 @@ def test_calc_lower_n_nan_ref(self):
...
 
+class TestCalcLowerE:
+ def create_input_df(self):
+        """Creates input df for test"""
+        input_cols = [
+            "employment",
+            "711",
+        ]
+        data = [
+            [1, 14],
```
adriano-lopresti: is there a reason why in column 711 you chose to put 14, which is also the sum of the column, i.e. the result of the function.. it is not wrong but it's a bit confusing, since one might also think what the function does is returning the "unique non-null entry of column 711", whilst instead it returns the sum of employment".

adriano-lopresti (commented):
@woodac , overall the function makes sense to me, but suggest a couple of small improvements

tests/test_estimation/test_calculate_weights.py
```diff
@@ -156,6 +156,69 @@ def test_calc_lower_n_nan_ref(self):
...
+        return input_df
+
+ def test_calc_lower_e(self):
+    """Test for calc_lower_e with nan."""
+    input_df = self.create_input_df()
+    expected_result = 14
+
+    # Call calc_lower_e function
+    actual_result = calw.calc_lower_e(input_df)
+    assert actual_result == expected_result, "calc_lower_e is behaving as expected"
```
AnneONS: this message should actually be what it prints if the test fails, so should say it's not working

tests/test_estimation/test_calculate_weights.py
```diff
@@ -156,6 +156,69 @@ def test_calc_lower_n_nan_ref(self):
...
+        ]
+        input_df = pd.DataFrame(data=data, columns=cols)
+        return input_df, cols, msg, emp_col
+
+    @pytest.mark.parametrize(
+        "cols, msg, emp_col",
+        [
+            (["employment", "711"], f"employment' or 711 missing.", "711"),
+        ],
+    )
```
AnneONS: can we talk through this? I am not fully familiar with using parametrize, but I'm not sure it's quite right even though it's passing.

tests/test_estimation/test_calculate_weights.py
```diff
@@ -156,6 +156,69 @@ def test_calc_lower_n_nan_ref(self):
...
+            [1, 10],
+            [4, np.nan],
+        ]
+        input_df = pd.DataFrame(data=data, columns=cols)
+        return input_df, cols, msg, emp_col
+
+    @pytest.mark.parametrize(
+        "cols, msg, emp_col",
+        [
+            (["employment", "711"], f"employment' or 711 missing.", "711"),
```
AnneONS: do we need an f-string here? And is that single quote meant to be there after employment?

AnneONS (commented):
the code looks fine, I'd just like to discuss the tests.

tests/test_estimation/test_calculate_weights.py
```diff
@@ -156,6 +156,63 @@ def test_calc_lower_n_nan_ref(self):
...
+
+    def test_calc_lower_e_missing_col(self):
+        """Test for calc_lower_e with missing col"""
+        input_df = self.create_input_df()
+
+        # Remove a required column to trigger an error
+        input_df = input_df.drop(columns=["711"])
+
+        # Apply function to see if error is raised
+        with pytest.raises(ValueError, match="employment' or 711 missing."):
```
AnneONS: we've still got the stray ' here

src/estimation/calculate_weights.py
```diff
@@ -40,6 +40,28 @@ def calc_lower_n(df: pd.DataFrame, exp_col: str = "709") -> dict:
...
+        df (pd.DatatFrame): The input dataframe which contains survey data,
+            including IDBR employment data.
+        emp_col (str): An appropriate column to store sum employment data.
+    Returns:
+        int: The sum of IDBR employment of sampled data within each cell.
+    """
+    # Check if any of the key cols are missing
+    cols = set(df.columns)
+    if not (("employment" in cols) & (emp_col in cols)):
+        raise ValueError(f"employment' or {emp_col} missing.")
```
AnneONS: have a look at the use of ' here

AnneONS (commented):
very nearly there, I think! Just a couple of style things now: can you put the 2 tests in the same class? And find a way to fix the stray '

AnneONS (approved):

PR #419, opened 2025-02-04
RDRP-1155_freezing_amendments_deletions

Ryan2Y79:
## Pull Request submission

Function added to src/freezing/freezing_apply_changes.py that removes subsequent instances when instance 0 has a 604 value of "No". Unit test accompanying added function. Furthermore, changes were necessary to already existing unit tests for them to pass (adding a 604 columns to the test dataframes). 

See ticket https://jira.ons.gov.uk/browse/RDRP-1155 for pipeline testing results.

#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/freezing/freezing_apply_changes.py
```diff
@@ -228,3 +232,58 @@ def apply_additions(
...
+            (accepted_amendments_df["instance"] == 0)
+            & (accepted_amendments_df["604"] == "No")
+        ]
+
+        # Get pairs of values for columns "reference" and "period"
+        # where instance = 0 and 604 = No.
+        reference_period_pairs = list(
+            zip(flagged_df["reference"], flagged_df["period"])
+        )
+
```
AnneONS, line 289: we don't need to deal with "period" column here- we only every have one period.

Ryan2Y79: comment has been addressed.

src/freezing/freezing_apply_changes.py
```diff
@@ -242,3 +246,59 @@ def apply_additions(
...
+    Returns:
+        pd.DataFrame: The main DataFrame with deletions applied
+    """
+
+    # Check if any row meets the condition of instance = 0 and 604 = No
+    condition_met = (
+        (accepted_amendments_df["instance"] == 0)
+        & (accepted_amendments_df["604"] == "No")
+    ).any()
```
AnneONS: inline with how we do things elsewhere in the pipeline, I'd do this, what do you think?

`    condition = (
        (accepted_amendments_df["instance"] == 0)
        & (accepted_amendments_df["604"] == "No")
    )

if condition.any():       # would this work?

flagged_df = accepted_amendments_df.loc[condition]

`

src/freezing/freezing_apply_changes.py
```diff
@@ -242,3 +246,59 @@ def apply_additions(
...
+        # Get pairs of values for columns "reference" and "period"
+        # where instance = 0 and 604 = No.
+        flagged_references = list(flagged_df["reference"])
+
+        # Iterate over the pairs and delete rows in main_df where
+        # instance is greater than 0
+        for reference in flagged_references:
+            main_df = main_df[
+                ~((main_df["reference"] == reference) & (main_df["instance"] > 0))
+            ]
```
AnneONS: `
reference_cond = df.reference.isin(flagged_referenced) & (df.instance > 0)`
`
main_df = main_df.loc[~reference_cond]
`

src/freezing/freezing_apply_changes.py
```diff
@@ -242,3 +246,59 @@ def apply_additions(
...
+
+        # Replace the value in column '604' with 'No' when reference is in
+        # flagged_references and instance is 0
+        main_df.loc[
+            (main_df["reference"].isin(flagged_references))
+            & (main_df["instance"] == 0),
+            "604",
+        ] = "No"
+    else:
```
AnneONS: can use reference_cond here as well

AnneONS (commented):
a few suggestions for you - good work though. 

Haven't looked at unit tests yet

src/freezing/freezing_apply_changes.py
Ryan2Y79: this works :D

src/freezing/freezing_apply_changes.py
Ryan2Y79: I think not as reference_cond is looking for instances greater than 0, however here we want instances equal to 0.

src/freezing/freezing_apply_changes.py
Ryan2Y79: I like this and have added this to the script. Cheers.

Ryan2Y79 (commented):
Thanks for the feedback. I much prefer the mask method having had to employ it. Cheers.

src/freezing/freezing_apply_changes.py
```diff
@@ -242,3 +246,60 @@ def apply_additions(
...
+        accepted_amendments_df["604"] == "No"
+    )
+
+    if condition.any():
+
+        # Filter the accepted_amendments_df Dataframe based on for
+        # instance =  0 and 604 = No
+        flagged_df = accepted_amendments_df[
+            (accepted_amendments_df["instance"] == 0)
+            & (accepted_amendments_df["604"] == "No")
```
AnneONS: how about putting `condition` in the square brackets?

src/freezing/freezing_apply_changes.py
AnneONS: good point!

AnneONS (commented):
another point for you... I'll just check your test.

tests/test_freezing/test_freezing_apply_changes.py
```diff
@@ -271,3 +272,54 @@ def test_apply_additions_no_amendments(self, frozen_df, dummy_additions, caplog)
...
+
+class TestApplyDeletions604(object):
+    """ Test for apply_deletions_604 """
+
+    def amendments(self) -> pd.DataFrame:
+        """A dummy amendments dataframe."""
+        columns = ["reference", "instance", "period", "510", "604"]
+        data = [
+            [4444, 0, 201812, 0.5, "No"],  # ref 4444 period 201812 to be flagged not removed
+        ]
```
AnneONS, line 309: I like the way you've added comments to explain what's being tested where

AnneONS (commented):
as discussed, we need to ensure all numeric values on instance 0 are re-set to zero

src/utils/breakdown_validation.py
```diff
@@ -22,40 +22,40 @@ def get_equality_dicts(config: dict, sublist: str = "default") -> dict:
     all_checks_dict = config["consistency_checks"]
 
     # isolate the relationships suitlable for checking in the construction module
-    if sublist == "default":
-        wanted_dicts = [
+    dict = {
```
AnneONS: I think dict is a keyword, so can you cange it?

AnneONS: Fab idea to do a dictionary though! much better.

src/utils/breakdown_validation.py
```diff
@@ -22,40 +22,40 @@ def get_equality_dicts(config: dict, sublist: str = "default") -> dict:
...
-        ]
+        ],
+        "604_check": ["4xx_totals", "5xx_totals", "7xx_a_totals", "7xx_b_totals"],
+    }
+
+    if sublist in dict.keys():
+        wanted_dicts = dict[sublist]
 
     else:
         wanted_dicts = list(all_checks_dict.keys())
```
AnneONS: We could change this now, cos I think that will retrun a list which a) contains duplicates, b) is not in any logical order.

The default is "default", so maybe we don't need this else statement which belonged to an earlier itteration of this code?

src/utils/breakdown_validation.py
```diff
@@ -22,43 +22,40 @@ def get_equality_dicts(config: dict, sublist: str = "default") -> dict:
...
+        "estimation": [
             "2xx_totals",
             "3xx_totals",
             "emp_xx_totals",
             "hc_xx_totals",
             "7xx_a_totals",
             "7xx_b_totals",
-        ]
+        ],
+        "604_check": ["4xx_totals", "5xx_totals", "7xx_a_totals", "7xx_b_totals"],
```
AnneONS: I wonder if it would be useful to rename this "instance_0_cols" in case it has more use?

src/utils/breakdown_validation.py
Ryan2Y79: Having looks at all instances where `get_equality_dicts` is called in the pipeline, the `else` statement is indeed redundant and unused. Good spot.

src/utils/breakdown_validation.py
Ryan2Y79: Done.

Ryan2Y79 (commented):
Comments addressed.

src/utils/breakdown_validation.py
Ryan2Y79: happy to rename sure

AnneONS (approved):

PR #432, opened 2025-02-17
RDRP-1171 new functions for filtering

AnneONS:
## Pull Request submission

The create_mask function in imputation_helpers.py previously had limited use as it doesn't work if not all the columns needed for all possible checks are in the dataframe.

For example, if a unit test doesn't include col "604", it's not possible to use create_mask at all.

A check that the correct column is in the dataframe is needed, and also, a way of calling each of the masks individually so we no longer need all possible

#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/utils/filter_helpers.py
```diff
@@ -0,0 +1,158 @@
+import pandas as pd
+from typing import List
+
+
+def create_notnull_mask(df: pd.DataFrame, col: str) -> pd.Series:
+    """Return a mask for string values in column col that are not null."""
+    return df[col].str.len() > 0
```
adriano-lopresti, line 7: With null here you mean that it's read in as an empty string, right?
I'm not sure what will happen if the column you are reading is a numeric one, and you have a null value.. 
maybe converting it into string will give you 'nan' or '<na>' the length of which is not zero?

Not saying is wrong, but worth being careful

src/utils/filter_helpers.py
```diff
@@ -0,0 +1,158 @@
...
+    msg = ""
+    for col in cols:
+        if col not in df.columns:
+            msg += f"Column {col} does not exist in this dataframe. "
+    if msg != "":
+        raise ValueError(msg)
+
+
+def get_clear_status_mask(df: pd.DataFrame) -> pd.Series:
+    check_cols_in_df(df, ["status"])
```
adriano-lopresti, line 29: General comment: if the column exists, this returns an error.
Didn't we said that you wanted also to have the possibility of applying the filter but with the option of avoiding the error of the column does not exist? in this case, whouldn't it make sense to add another boolean argument called, say, "error_if_no_column", which, if a column doesn't exist, makes the "check_cols_in_df" skip the filter rather than throwing an error?

Or did you want to avoid an error if even a column on which you were not filtering, didn't exist? In this case, I think you are implementing it correctly.

src/utils/filter_helpers.py
adriano-lopresti: PS: I see from the unit test that it seems to work, so I am ok with it..
But my comment still stands as something to be careful in case all seems to look fine in the unit tests but then when real "edge-case" data are used then it suddenly doesn't behave as expected

tests/test_utils/test_filter_helpers.py
```diff
@@ -0,0 +1,230 @@
...
+        result_mask = create_mask(df, options)
+        assert_series_equal(result_mask, expected_mask, check_names=False)
+
+    def test_not_mor_imputed_longform(self, create_input_df):
+        df = create_input_df
+        options = ["not_mor_imputed", "longform_only"]
+        expected_mask = pd.Series([False, False, True, True, True, False])
+        result_mask = create_mask(df, options)
+        assert_series_equal(result_mask, expected_mask, check_names=False)
+
```
adriano-lopresti, line 199: I'm happy with the tests so far.
You still seem to be working to the test below.

adriano-lopresti (commented):
all good. I just have a couple of comments

src/utils/filter_helpers.py
AnneONS: This function was introduced by GZ who was looking at all the different ways nulls work. Maybe we should change its name so we know to only use it for string cols?

src/utils/filter_helpers.py
AnneONS: it's the latter, and the way the other functions are set up will allow us to do everything we want to.
I think we really do want to throw an error if the code tells us to filter on a column and it doesn't exist.

AnneONS (commented):
Here are my responses to your questions @adriano-lopresti  
thanks for the review.

PR #433, opened 2025-02-18
Rdrp 1159 calculate growth rates unit test

abaizan:
## Pull Request submission

- Wrote a unit test for calculate_growth_rates
- changed unit test helper to take multiple files

#### Closes or fixes
[RDRP-1159](https://jira.ons.gov.uk/browse/RDRP-1159)
Unit test for calculate_growth_rates func


tests/test_imputation/test_MoR.py
```diff
@@ -168,3 +169,148 @@ def test_run_mor_short_form(
...
+        [1047, 1, 400, 0, 0, "R", "C_BC", "C"],
+        [1047, 2, 200, 0, 0, "R", "D_BC", "C"],
+    ]
+
+        test_CGR_current_df = pd.DataFrame(data=data, columns=test_CGR_current_columns)
+        return test_CGR_current_df
+
+    @pytest.fixture(scope="function")
+    def create_test_CGR_backdata_df(self):
+        """Create an test_CGR_backdata dataframe for the test."""
```
woodac, line 260: Everything looks good to me. All I picked up on was there are little red squiggles are the start of some of the lines, due to indention being slightly out. Maybe you could ask co-pilot to correct these for you, to save you the time going through and removing them ? :)

tests/test_imputation/test_MoR.py
```diff
@@ -168,3 +169,148 @@ def test_run_mor_short_form(
...
+        target_vars_list
+
+    ):
+        """Test the calculate_growth_rates function."""
+        current_df = create_test_CGR_current_df
+        backdata_df = create_test_CGR_backdata_df
+        expected_df = create_test_CGR_expected_df
+        target_vars = target_vars_list
+
+        result_df = calculate_growth_rates(current_df, backdata_df, target_vars)
```
woodac, line 324: Worked out calculation myself and got the same results.

helpers/unit_test_helper.py
```diff
@@ -1,93 +1,98 @@
...
 import os
 
-# configuration settings
-csv_path = "D:/coding_projects/randd_test_data/"
-input_file = "weights2_input.csv"
+filename_string_cols_dict = {
+    "test_CGR_current.csv" : ["200", "201", "601", "604", "status", "formtype", "postcodes_harmonised", "imp_marker", "imp_class", "selectiontype"],
+    "test_CGR_backdata.csv": ["200", "201", "601", "604", "status", "formtype", "imp_marker", "imp_class", "selectiontype"], 
+    "test_CGR_expected.csv": ["imp_class"],
+}
```
AnneONS, line 11: the idea here is to create a class with methods to create input and expected dfs.
We want a df called "input_df" or "expected_df", and functions "create_input_df" etc.
So you might like to rename your input files to not have "test", but to match the generic function names like "input..."

helpers/unit_test_helper.py
```diff
@@ -1,93 +1,98 @@
...
 
-# whether the unit test data is input or expected output (set "input" or "exp_output")
-in_or_output = "input"
+for filename, string_cols in filename_string_cols_dict.items():
+    # configuration settings
+    csv_path = "D:/RDSA_R&D/"
```
AnneONS: this is different for all of us, of course, so could it go right at the top, along with where we enter the filenames?

helpers/unit_test_helper.py
```diff
@@ -1,93 +1,98 @@
...
+    input_file = filename
 
-output_filename = f"{in_or_output}_function"
+    # whether the unit test data is input or expected output (set "input" or "exp_output")
+    in_or_output = input_file.split(".csv")[0]
```
AnneONS: here we want to return "input", "output", "expected" etc, hence my comment above on your  choice of csv names.

tests/test_imputation/test_MoR.py
```diff
@@ -168,3 +169,148 @@ def test_run_mor_short_form(
         assert_frame_equal(result_df, expected_sf_mor_output, check_dtype=False, check_exact=False), (
             "run_mor() not imputing data as expected."
         )
+
+class Test_calculate_growth_rates(object):
+    """Tests for calculate_growth_rates."""
+
+    @pytest.fixture(scope="function")
```
AnneONS: you don't need to use @pytest.fixture here within the function- you can just refer to it using self.target_vars_list() 

@pytest.fixture is useful to be able to access input data inside more than one function, or more than one class.

tests/test_imputation/test_MoR.py
```diff
@@ -168,3 +169,148 @@ def test_run_mor_short_form(
...
+        "selectiontype",
+    ]
+
+        data = [
+        [1031, 1, 0, 10, 0, "R", "C_AA", "C"],
+        [1031, 1, 0, 10, 0, "R", "C_AA", "C"],
+        [1031, 1, 0, 10, 0, "R", "C_AA", "C"],
+        [1031, 2, 0, 0, 0, "R", "D_AA", "C"],
+        [1031, 2, 0, 0, 0, "R", "D_AA", "C"],
+        [1031, 2, 0, 0, 0, "R", "D_AA", "C"],
```
AnneONS: this duplication thing in theory shouldn't happen- it might just be a rare edge case. I suggest we don't need to have so many lines testing the same thing, just a couple of cases to check the groupby works as expected.

tests/test_imputation/test_MoR.py
```diff
@@ -168,3 +169,148 @@ def test_run_mor_short_form(
...
+
+
+    def test_calculate_growth_rates(
+        self,
+        create_test_CGR_current_df,
+        create_test_CGR_backdata_df,
+        create_test_CGR_expected_df,
+        target_vars_list
+
+    ):
```
AnneONS: if you don't use the @fixture, you could do this instead:
def test_calculate_growth_rates(self):
    create_test_CGR_current_df = self.create_test_CGR_current_df() or whatever works.

tests/test_imputation/test_MoR.py
```diff
@@ -168,3 +169,148 @@ def test_run_mor_short_form(
...
+        "emp_technician_gr",
+    ]
+
+        data = [
+        [1031, "C_AA", 0, 30, 0, 0.0, 10.0, 0.0, np.nan, 3.0, np.nan],
+        [1031, "D_AA", 0, 0, 0, 0.0, 0.0, 0.0, np.nan, np.nan, np.nan],
+        [1045, "C_AH", 0, 0, 30, 10000.0, 0.0, 10.0, np.nan, np.nan, 3.0],
+        [1045, "D_AH", 0, 0, 30, 10000.0, 0.0, 10.0, np.nan, np.nan, 3.0],
+        [1047, "C_BC", 400, 0, 0, 400.0, 0.0, 0.0, 1.0, np.nan, np.nan],
+        [1047, "D_BC", 200, 0, 0, 200.0, 0.0, 0.0, 1.0, np.nan, np.nan],
```
AnneONS: I think I'd like to see a few more non-null growth rates. 
The cases to cover are:
current and prev both not null and non-zero
one of the two is null
one of the two is zero

AnneONS (commented):
some very good stuff! A few comments- hope you can put up with me telling you comments in line with our "house style". Thanks.

helpers/unit_test_helper.py
abaizan: Hi @AnneONS , I just want to check - would changing "test_CGR_current.csv" > "CGR_current_df.csv" match the style enough? I'm confused by the "input_df" as I had to make both current and backdata files, but I might be thinking about it too literally

AnneONS: yes, sure. Personally I wouldn't feel the need to "namespace" with CGR if it's inside a class, but up to you!

tests/test_imputation/test_MoR.py
abaizan: Ah yes this table had a lot more columns before that could have had different values. They don't make sense now, I'll change it

AnneONS (commented):
another comment- we'll chat more soon

AnneONS (approved):
This is really great! Very good test.

AnneONS (approved):

PR #434, opened 2025-02-18
RDRP-1172_construction

Ryan2Y79:
## Pull Request submission

Construction hotfix. The addition of this line in a previous PR was totally redundant, and did not have the intended effect as to whether NI construction was run or not. Furthermore, the addition of this line is causing the pipeline to die during construction. 

#### Closes or fixes

Closes #

#### Code

- [x] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)

### DIFF
--- a/src/pipeline.py
+++ b/src/pipeline.py
@@ -162,7 +162,6 @@ def run_pipeline(user_config_path, dev_config_path):  # noqa C901
             mods.rd_file_exists,
             mods.rd_read_csv,
             is_run_all_data_construction=True,
-            is_northern_ireland=config["global"]["run_ni_construction"],
         )
     else:
         MainLogger.info("All data construction is not enabled")


adriano-lopresti (commented):
i just had a quick look at it, but my questions are:
1) it's just one line you've added?
2) what you added with that line, is adding one argument when the function run_constructions is called.. and i notice that the function is already defined with that argument (but it's an optional value, the default being False)..
3) i see that, in the config, the value of that parameter is False, which is the same as the default parameter
 
so the question is: what is the context of this? it seems that effectively, with this change nothing as changed.
Except that it gives the possibility of switching that parameter from False to True. Have you tested what happen if this is switched to True? I checked that the pipeline still runs through, but i haven't verified the impact of doing so.
 
But i still find the logic of it all a bit puzzling
(but what i find puzzling was actually there before your PR...your PR gives the possibility of "switching off" this chunk of code.. so maybe one should jsut remove this line 56..)
What i find puzzling is: to me is seems that the variable run_construction defined in line line 56 of contruction_main is redundant 
You pass a "is_northern_ireland" argument(which is read from the run_ni_contruction config parameter).
Then if this true, you define a variable run_construction being equal to the same config parameter (why?) which doesn't seem to be subsequenrly used.

Ryan2Y79:
The change I prepose here is reversing a line I added in this PR (that you approved and merged) https://github.com/ONSdigital/research-and-development/pull/409/files. 

The overall context is that Northern Ireland and GB construction should happen independently. Norther Ireland construction happens `pipeline.py` lines 139-153, and GB construction happens in `pipeline.py` lines 155-168. However BOTH construction executions use the same construction function `run_construction`. 

The reason for the `is_northern_ireland` parameter and its default value to set to `False` is that when GB construction is run, we do NOT want NI construction to happen. And when GB constructions happens, we do NOT want NI construction to happen. Both GB and NI construction should happen happen independently. My change in PR `https://github.com/ONSdigital/research-and-development/pull/409/files` violates this idea by having Northern Ireland construction happen at the same time as GB if config["global"]["run_ni_construction"] is set to `True`

You are totally correct in that with the config["global"]["run_ni_construction"] set to False, my change here makes no difference. But when config["global"]["run_ni_construction"] is set to True, the pipeline will die as we have introduced NI construction during GB construction. 

In short, I want to reverse a change that I made previously to restore the pipeline to how it once was.

adriano-lopresti (approved):
I approve

PR #438, opened 2025-02-25
RDRP-1175_ITL_outputs_additional_cols

woodac:
Additional HC and FTE employment columns to ITL1 and 2 GB and UK outputs

#### Closes or fixes

Closes #

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


Ryan2Y79:
You are totally correct that the UK and GB ITL 1&2 outputs are the same, which makes no sense as UK includes Northern Ireland, and GB does not include Norther Ireland. Therefore we should expect a difference between the UK and GB outputs.

However my thinking is this.... there are two unit tests, `test_aggregate_itl_gb` and `test_aggregate_itl_uk` which are both passing currently. The expected output for both of these unit test are exactly the same! How is this if one of the unit tests is for GB and one for UK?! The expected dataframes for `test_aggregate_itl_gb` and `test_aggregate_itl_uk` should be different. One of those unit tests has not been constructed properly to test the `uk_output` parameter pertaining to the function `aggregate_itl`. 

My suggestion is 1. redefine at least one of the expected dataframes for either `test_aggregate_itl_gb` and/or `test_aggregate_itl_uk` so that the unit test is actually doing a proper test. 2. The unit test should then fail, and 3. find out why UK and GB are producing the same outputs. Make the unit test your barometer as to whether you have fixed the UK/GB issue.

src/outputs/intram_by_itl.py
```diff
@@ -94,27 +94,41 @@ def aggregate_itl(
...
+        for col in geo_cols + ["postcodes_harmonised"]:
             ni_df[col] = pd.NA
-        df = df.append(ni_df, ignore_index=True).copy()
+        df = pd.concat([df, ni_df], ignore_index=True).copy()
+    # Drop any NI data that is in existing df
+    else:
+        for col in geo_cols:
+            # Remove NI data from the GB data
+            df = df[~df[col].str.contains("TLN", na=False)]
+
```
woodac, line 113: "Else" statement added, as the ni_full_reponses is passed into the function, so regardless of the bool being true or false there was always NI data in the original df. Else statement drops any lingering Northern Ireland data.

woodac: GB output should have NI data, as there could be offices in NI for GB, the figure data should differ between the two outputs.

src/user_config.yaml
Ryan2Y79, line 1: I'd suggest not committing user config changes to develop.

woodac (commented):
I have run and checked the output figures which look much more accurate and logical to me for the gb/ uk split.  I can't see an issue with code and tests, they look good to me.

Ryan2Y79 (commented):
All looks logical to me. Please address comment on `user_config,py` changes.

src/user_config.yaml
AnneONS: ok good point, I'll take out the NI one but the other is a correction = it should be false

Ryan2Y79 (approved):
All good!

PR #447, opened 2025-03-12
RDRP-1181_NA_output_PNP

woodac:
National Accounts require an output in the same format as the old 2021 PNP data.

This will involve mapping questions from the output after apportionment.

Steps to do this:

1. the 2xx and 3xx questions will need to be mapped to corresponding questions, with C and D moving from separate rows to separate cols.

2. Division by 1000.

3. TO_DO- Don't map 4xx and 5xx columns, instead ignore these and map headcount_xx and emp_xx columns.

The columns on address etc at the beg of the file are different to ours- don't worry so much about these, get the numeric qs done.

Link to our previous correspondence guide:

https://officenationalstatistics.sharepoint.com/:x:/r/sites/ESCSEPDS/_layouts/15/Doc.aspx?sourcedoc=%7BBBD8C956-3C83-4BA5-8D5A-D74C518FFAFE%7D&file=PNP%20questions%20compared%20with%20BERD%20ANSWERS.xlsx&action=default&mobileredirect=true

Notes:

Use src.outputs.frozen_group.py, which does similar for BERD, as a guide, you may need to do some of the same steps.

in the helpers.create_pnp_backdata.py script, use the mappings of old to new questions, and look for where we do summing. Look for cols that we previously deleted or created, we may need to do the reverse.

#### Code

- [ ] **Clean Code**

#### Documentation

#### Data

#### Testing

---

# Peer Review Section

---

### Review comments

  works correctly?)


src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
+"""National Accounts Output for PNP"""
```
AnneONS: can the file be renamed PNP_NA_output.py?

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+OutputMainLogger = logging.getLogger(__name__)
+
+
+def divide_by_1000(df, config):
+    """Values in columns starting 2xxx or 3xxx are divided by 1000"""
+    # Get cols from config
+    cols = get_all_wanted_columns(config, "imputation")
+
+    for col in cols:
+        df = df.copy()
```
AnneONS: we don't need to do df = df.copy() evey time we go through one of the columns- this line should go outside the for loop

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def expenditure_by_region(df):
+    """Calculation to populate the expenditure by region column."""
+    # Add column for expenditure by region
+    df["Expenditure by Region"] = (
+        (df["602"].fillna(0)) / 100 * (df["211_C"].fillna(0) + df["211_D"].fillna(0))
+    )
+    df["Expenditure by Region"] = df["Expenditure by Region"].astype(float).round(2)
+    return df
```
AnneONS: I don't think we need to do this- we have proportioned everything by region in the regional_apportionment module, and all our responses have already been multiplied by column 602

Also, this calc wouldn't be right, as the percentages in col 602 don't "line up" with the values in col 211. We might have 3 instances with a 211 value, but 2 postcodes. Regional apportionment takes care of all this.

We need to get back to Jim to discuss this: their "region" is the value in IDBR, so the same for all rows on the same RU. So the default would be just to copy the 211 value here and explain.

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def remove_C_D(df: pd.DataFrame):
+    """Removes _C or _D from columns where it is unnecessary"""
+
+    # Remove _C or _D from columns that do not start with a number
+    for col in df.columns:
+        # Doesn't start with a number
+        if not col[0].isdigit():
+            df = df.rename(columns={col: col[:-2]})
```
AnneONS: Have you thought of only creating the _C and _D versions where it IS necessary?

Note: We do need to do _C and _D for the emp_ and headcount_ cols

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Remove _C or _D from 6XXX cols
+    # Get 6XXX cols
+    cols = [col for col in df.columns if col.startswith("6")]
+    for col in cols:
+        df = df.copy()
+        if col in df.columns:
+            df = df.rename(columns={col: col[:-2]})
+
```
AnneONS: can we not put in the _C and _D to start with? Remember that we don't need to apply the percentage 602 as this was already done.

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+
+    # Check that PNP is set
+    if config["survey"]["survey_type"] != "PNP":
+        raise ValueError(
+            "The survey type is not set to PNP. "
+            "The national accounts output is for PNP only "
+            "and will not be created."
+        )
```
AnneONS: Instead of having to do a valueError here, which will stop the pipeline, could you have logic in outputs_main.py which means we only come to this script if it's a PNP run?

src/outputs/outputs_main.py
```diff
@@ -219,4 +220,10 @@ def run_outputs(  # noqa: C901
         output_intram_totals(intram_tot_dict, config, write_csv)
         OutputMainLogger.info("Finished Intramural totals output.")
 
+    # Running National Accounts output
+    if config["global"]["output_na"]:
```
AnneONS: can you add in the condition on the survey_type as well here?

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+    cols = [col for col in df.columns if col.startswith("6")]
+    for col in cols:
+        df = df.copy()
+        if col in df.columns:
+            df = df.rename(columns={col: col[:-2]})
+
+    return df
+
+
+def output_na(df: pd.DataFrame, config: dict, write_csv: callable):
```
AnneONS: What do you think about renaming this output_pnp_na so it's clearer when we call it in outputs_main.py what this output is for?

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+    # Select only the columns we need
+    div_df = df.copy()
+    div_df = div_df[cols]
+
+    # Divide by 1000
+    div_df = divide_by_1000(div_df, config)
+
+    # Replace cols in df with the new values
+    df[cols] = div_df[cols]
+
```
AnneONS: I think it would be better not to create a div_df. Just work with df. You are doing .copy() in the divide_by_1000 function

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+    # Filter civil and defence data into seperate dataframes
+    civil_df = df[df["200"] == "C"]
+    defence_df = df[df["200"] == "D"]
+
+    # Add "C" or "D" to the columns of the corresponding dataframes
+    civil_df.columns = civil_df.columns + "_C"
+    defence_df.columns = defence_df.columns + "_D"
+
```
AnneONS: how about if you do this: 

```suggestion
# Add "C" or "D" to the columns of the corresponding dataframes
# 2xx, 3xx, emp_xx, headcount_xx
wanted_cols = get_all_wanted_columns(config, "long_forms")
civil_df = civil_df.rename(columns={col: col + "_C" for col in wanted_cols})
defence_df = defence_df.rename(columns={col: col + "_D" for col in wanted_cols})

```

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+    # Remove _C or _D from columns that do not start with a number or 6XX cols
+    civil_df = remove_C_D(civil_df)
+    defence_df = remove_C_D(defence_df)
```
AnneONS: would this be unnecessary if we do the suggestion above?

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Concatenate the dataframes
+    df = pd.concat([civil_df, defence_df], ignore_index=True, axis=0)
```
AnneONS: we now need to do a join, see ticket 1205
join_cols = ["reference", "201", "601"]

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Map to the CORA statuses from the statusencoded column
+    df = create_cora_status_col(df)
+
+    # Calculate the expenditure by region
+    df = expenditure_by_region(df)
```
AnneONS: Pretty sure we don't need this- it's already done.

config/output_schemas/national_accounts_schema.toml
```diff
@@ -0,0 +1,983 @@
+[InquiryIDBRCode]
+name = ""
+old_name = "survey"
+Deduced_Data_Type = "str"
+R_and_D_Type = "survey"
```
AnneONS: do you think a better name for "R_and_D_Type" might be "temp_name" ? cos we're creating these temporary column names.

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+    # Create dict of current and required column names
+    colname_schema_dict = {
+        output_schema[column_nm]["R_and_D_Type"]: column_nm
+        for column_nm in output_schema.keys()
+    }
+
+    # If col is in schema but not in df, add it to the df
+    for col in colname_schema_dict.keys():
+        if col not in df.columns:
+            df = df.copy()
```
AnneONS: again, we don't need to keep doing df = df.copy() every time we go through the loop. Perhaps we don't need it here at all

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Create a DataFrame for the first row with the original column names
+    first_row_values = list(colname_schema_dict.values())
+    first_row_df = pd.DataFrame([first_row_values], columns=new_column_names)
+
+    # Rename the columns of the output DataFrame to the new column names
+    output_df.columns = new_column_names
+
+    # Concatenate the first row DataFrame with the output DataFrame
+    output_df = pd.concat([first_row_df, output_df], ignore_index=True)
```
AnneONS: can we have a separate function that creates this special row?

AnneONS (commented):
Lots of work here- you've been busy! Main changes: we need to do _C and _D versions for headcount and emp_ cols as well. Also, no need to do percentage by region.

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Select only the columns we need
+    div_df = df.copy()
+    div_df = div_df[cols]
+
+    # Divide by 1000
+    div_df = divide_by_1000(div_df, config)
+
+    # Replace cols in df with the new values
+    df[cols] = div_df[cols]
```
AnneONS: this line makes me a bit nervous! I know what you're trying to do, but I have a feeling this doesn't always work the way we expect it to in all versions of python. See comment below

src/outputs/NA_output.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Filter civil and defence data into seperate dataframes
+    civil_df = df[df["200"] == "C"]
+    defence_df = df[df["200"] == "D"]
+
+    # Add "C" or "D" to the columns of the corresponding dataframes
+    civil_df.columns = civil_df.columns + "_C"
+    defence_df.columns = defence_df.columns + "_D"
```
AnneONS: Change of plan- see new ticket 1205- we need to merge civil_df and defence_df at the end. So for this to work, we need one of them, say defence_df, to just have the cols we need:
```
# 2xx, 3xx, emp_xx, headcount_xx
wanted_cols = get_all_wanted_columns(config, "long_forms")
join_cols = ["reference", "201", "601"]
defence_df = df[df["200"] == "D"][join_cols + wanted_cols]
```

tests/test_outputs/test_national_accounts.py
```diff
@@ -0,0 +1,145 @@
...
+                    "emp_defence": []
+                }
+            }
+        }
+        return config
+
+    def input_data(self):
+        """Input data for divide_by_1000 tests."""
+        columns = ["212", "250", "247"]
+        data = [[1, 500, 1000], [2, 1000, 2000], [3, 2000, 3000]]
```
AnneONS: can we have the component lists on separate lines so it looks like a mini dataframe? As it's a test, it's ignored by the precommit hooks.

tests/test_outputs/test_national_accounts.py
```diff
@@ -0,0 +1,145 @@
...
+        df = pd.DataFrame(data=data, columns=columns)
+        return df
+
+    def expected_output(self):
+        """Expected output data for divide_by_1000 tests."""
+        columns = ["212", "250", "247"]
+        data = [[0, 0, 1], [0, 1, 2], [0, 2, 3]]
+        df = pd.DataFrame(data=data, columns=columns)
```
AnneONS, line 48: interesting!! Does python round 0.5 down? 
I just asked copilot and I learnt something:
"In Python, the built-in [round()] function rounds 0.5 up to the nearest even number. This behavior is known as "round half to even" or "bankers' rounding". It is the default rounding mode in Python to reduce cumulative rounding errors in calculations."
0.5 is rounded to 0 because 0 is the nearest even number.
1.5 is rounded to 2 because 2 is the nearest even number.
2.5 is rounded to 2 because 2 is the nearest even number.
3.5 is rounded to 4 because 4 is the nearest even number.

I never knew!

tests/test_outputs/test_national_accounts.py
```diff
@@ -0,0 +1,145 @@
...
+
+        assert_frame_equal(
+            result_df.reset_index(drop=True),
+            expected_output.reset_index(drop=True),
+            check_dtype=False
+        )
+
+
+class TestExpentitureByRegion(object):
+    """ Test to check the generation and calculation of expenditure by region col."""
```
AnneONS: sorry, I don't think we need this

AnneONS (commented):
a few more things to work at - I like your comprehensive approach to testing.

tests/test_outputs/test_national_accounts.py
woodac: Removed :)

src/outputs/PNP_NA_output.py
```diff
@@ -0,0 +1,145 @@
...
+OutputMainLogger = logging.getLogger(__name__)
+
+
+def divide_by_1000(df, config):
+    """Values in columns starting 2xxx or 3xxx are divided by 1000"""
+    # Get cols from config
+    cols = get_all_wanted_columns(config, "imputation")
+
+    for col in cols:
+        df = df.copy()
```
AnneONS: we don't want to have this line every time we go through the loop. You could put it at the beginning of the function outside the loop.

src/outputs/PNP_NA_output.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Divide by 1000
+    df = divide_by_1000(df, config)
+
+    # Get the columns that are required for the output
+    wanted_cols = get_all_wanted_columns(config, "longform")
+
+    #  Add "C" or "D" to the columns of the corresponding dataframes
+    civil_df = df.rename(columns={col: col + "_C" for col in wanted_cols})
+    defence_df = df.rename(columns={col: col + "_D" for col in wanted_cols})
```
AnneONS: we still need to do the conditions on df.loc[df["201"] == "C/D"]

src/outputs/PNP_NA_output.py
```diff
@@ -0,0 +1,145 @@
...
+    # Get columns consistent in both dataframes
+    join_cols = ["reference", "201", "601"]
+
+    # Merge the dataframes
+    join_df = pd.merge(
+        civil_df, defence_df, on=join_cols, how="outer", suffixes=("", "_drop")
+    )
+
+    # Drop the columns that are not needed
+    join_df = join_df.loc[:, ~join_df.columns.str.endswith("_drop")]
```
AnneONS: instead of doing this, you could create the second df, defence_df, so that it only contains [join_cols + wanted cols]

AnneONS (commented):
it's getting there! A few more comments for you

AnneONS (commented):
coming on great!

src/outputs/PNP_NA_output.py
woodac: Ah yes, I think I did this because I was getting a pandas setting copy warning and this stopped it. I'll move it outside the loop because I know we have a ticket for the copy warning setting

src/outputs/PNP_NA_output.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # Divide by 1000
+    df = divide_by_1000(df, config)
+
+    # Get the columns that are required for the output
+    wanted_cols = get_all_wanted_columns(config, "longform")
+
+    # Filter civil and defence data into seperate dataframes
+    civil_df = df[df["200"] == "C"]
+    defence_df = df[df["200"] == "D"]
```
AnneONS: maybe add a .copy() here? ie, = df.copy().[..........

src/outputs/PNP_NA_output.py
```diff
@@ -0,0 +1,146 @@
...
+
+    #  Add "C" or "D" to the columns of the corresponding dataframes
+    civil_df = civil_df.rename(columns={col: col + "_C" for col in wanted_cols})
+    defence_df = defence_df.rename(columns={col: col + "_D" for col in wanted_cols})
+
+    # Get columns consistent in both dataframes
+    join_cols = ["reference", "201", "601"]
+    defence_df = defence_df[join_cols + [col + "_D" for col in wanted_cols]]
```
AnneONS: this line here, which I suggested, is causing your bug, as it means any rows which are defence only don't have all the wanted column info.

Think it possibly wouldn't work to go back to your previous method of joining the two full dataframes, and dropping the rows which are duplicate, I think we might need to do the following:

civil_df = civil_df[join_cols + [col + "_C" for col in wanted_cols]]
defence_df = defence_df[join_cols + [col + "_D" for col in wanted_cols]]

merge_df = pd.merge(civil_df, defence_df, on=join_cols, how="outer")

now merge this back onto the original df- hopefully it will work if none of the cols in merge_df are the same as any in the original df except join_cols

AnneONS (commented):
it's getting there! Nearly finished now I think

src/outputs/PNP_NA_output.py
```diff
@@ -0,0 +1,146 @@
...
+    Args:
+        df (pd.DataFrame): Dataframe containing all columns
+        output_schema (str): Toml schema containing the old and new
+        column names for the outputs
+
+    Returns:
+        (pd.DataFrame): A dataframe consisting of only the
+        required short form output data
+    """
+    # Create dict of current and required column names
```
AnneONS: I can see this code is available in this function create_output_df in output_helpers- but we are using a different column then old_name. Do you think we could modify the function create_output_df so has an extra optional variable from_name defaults to "old_name" but we can use "R_and_D_type" ?

src/outputs/PNP_NA_output.py
```diff
@@ -0,0 +1,146 @@
...
+    output_df.rename(columns=colname_schema_dict, inplace=True)
+
+    # Rearrange to match user defined output order
+    output_df = output_df[colname_schema_dict.values()]
+
+    # Create a list of new column names using the "name" field from the schema
+    new_column_names = []
+    for col in colname_schema_dict.values():
+        new_column_names.append(output_schema[col]["name"])
+
```
AnneONS: how about this:
    description_schema_dict = {
        output_schema[column_nm]["name"]: column_nm
        for column_nm in output_schema.keys()
    }
and now hopefully from this dictionary, which includes nulls, can you get the following?

description_row = list(description_schema_dict.keys())

AnneONS (commented):
just a suggestion or two for the last part, but I understand the code now :-)

src/outputs/outputs_main.py
```diff
@@ -219,4 +220,10 @@ def run_outputs(  # noqa: C901
         output_intram_totals(intram_tot_dict, config, write_csv)
         OutputMainLogger.info("Finished Intramural totals output.")
 
+    # Running National Accounts output
+    if config["global"]["output_pnp_na"]:
+        OutputMainLogger.info("Starting PNP National Accounts output...")
```
AnneONS, line 225:
```suggestion
    if config["global"]["output_pnp_na"] and config["survey"]["survey_type"] == "PNP":
```

src/outputs/PNP_NA_output.py
```diff
@@ -0,0 +1,99 @@
...
+    cols = get_all_wanted_columns(config, "imputation")
+
+    for col in cols:
+        if col in df.columns:
+            df[col] = df[col].apply(lambda x: round(x / 1000, 0) if x > 0 else x)
+
+    return df
+
+
+def output_pnp_na(df: pd.DataFrame, config: dict, write_csv: callable):
```
AnneONS, line 61: This function is the "entry point" function for the script, we usually have it at the bottom of the script, it helps us navigate. Can you move this to be after the function below? Thanks

src/outputs/PNP_NA_output.py
```diff
@@ -0,0 +1,99 @@
...
+
+def divide_by_1000(df, config):
+    """Values in columns starting 2xxx or 3xxx are divided by 1000"""
+    # Get cols from config
+    cols = get_all_wanted_columns(config, "imputation")
+
+    for col in cols:
+        if col in df.columns:
+            df[col] = df[col].apply(lambda x: round(x / 1000, 0) if x > 0 else x)
+
```
AnneONS, line 26: I asked copilot if we could do this without the loop- it's good to do things in a vectorised way.

```suggestion
    # Ensure only columns that exist in the DataFrame are processed
    cols_to_process = [col for col in cols if col in df.columns]

    # Apply the transformation to all relevant columns
    df[cols_to_process] = df[cols_to_process].applymap(lambda x: round(x / 1000, 0) if x > 0 else x)

```

AnneONS (commented):
nearly ready now!

AnneONS (commented):
This is nearly ready to approve, just one or two suggestions/ requests.

AnneONS (approved):

